PR #18, opened 2018-05-02
adds option to remove get started button

aaronbuchwald:
search options.getStarted === "false" to see the four lines of code added besides formatting changes
to test add option getStarted: false


monty0157:
So, the merge conflict is pretty intense, and I will therefore just add the lines by hand. But good work!

PR #19, opened 2018-05-11
changes css for gallery subtitle and title div

aaronbuchwald:
fixes the issue where when there is no title and subtitle it still renders a large blank div


monty0157:
The height of the container is unfortunately necessary in order to make all carousel items have the same height :)

PR #25, opened 2018-07-19
Fixed compatibility for older browsers

dshire:
Fixed browser compatibility for IE11: 

- styling (flexbox compatibility)
- colors
- polyfill for fetch
- workaround for TTS and STT


src/index.js
```diff
@@ -139,8 +149,21 @@ async function init(userOptions: any, outputCallback: (output: { text: string, d
...
+
+		const bgColorCollection = document.querySelectorAll(".button_template_text,	.cognigy-chat-state-closed,	.cognigy-chat-state-open, .cognigy-chat-header-container, .cognigy-chat-header-container__open, .cognigy-chat-bot-message");
+		changeColor(bgColorCollection, "background-color", options.colorScheme);
+
+		function changeColor(coll, key, value){
+			for(var i=0, len=coll.length; i<len; i++)
+			{
+				coll[i].style[key] = value;
+			}
+		}
```
monty0157: Do we still need this?

src/index.js
```diff
@@ -3,6 +3,11 @@
 
 /* Node modules */
 import { h, render } from 'preact';
+import 'whatwg-fetch';
+import Promise from 'promise-polyfill';
+if (!window.Promise) {
+		window.Promise = Promise;
+	}
```
monty0157: Linting :D

monty0157: The colors on the "facebook" elements are not set - so e.g for buttons on the carousel

monty0157: image

src/index.js
monty0157: No logos for most messages

image

monty0157:
image

PR #29, opened 2018-08-14
Fix null message rendering

j-sommer:
**Bug**
Messages with an empty string ("") or of type **null** are rendered as an empty speech bubble.
screenshot from 2018-08-14 11-25-06
**Fix**
Edited check to skip rendering all messages where the message text is a falsey value.


the2ndlaw:
LGTM!

monty0157:
Looks good

PR #31, opened 2018-11-06
Force websockets for Safari

monty0157:
We now force websockets for Safari to deal with cookie based sticky sessions with Traefik. Safari doesn't send cookies with requests when the requested domain is different, meaning that polling will go to different endpoint instances since the cookie used sticky sessions is missing. We can solve this by forcing websockets for Safari.

**Reproducing the issue**
Testing and reproducing this issue is sadly a bit complex, so you need to have a bit of patience. In order to reproduce this, you need to:

- Run 3 replicas of the Endpoint service
- Create a entry in your /etc/hosts file mapping 127.0.0.1 to webchat.cognigy.com
- Manually change the Traefik Host rule for the Webchat in your cognigy.yml file. Change it from Host:webchat to Host:webchat.cognigy.com
- Open http://webchat.cognigy.com/<your-url-token> in Safari. You can either do this on a Mac or use BrowserStack to launch Safari version 12 on Mojave. Our login credentials are in the wiki.
- You will not receive any messages in the Webchat. Open the Safari developer tools in the develop menu bar. If the develop menu bar isn't shown, then go to Preferences > advanced > Show Develop menu in menu bar.
- Go to the network panel and you will see that socket.io isn't able to connect.

**How to test**
After reading the code, follow the steps above to setup a testing environment. Checkout the corresponding branch in the cognigy repo and run the stack. This will give you a new service-webchat that includes the minified file from this PR. Open the webchat in Safari once again, and you should see that it connects via websockets immediately.


mayrbenjamin92:
lgtm

PR #34, opened 2018-11-22
Gallery Element OnClick Handling

j-sommer:
Opens the default action web url (when defined) or the image in a new tab


mayrbenjamin92:
LGTM

PR #38, opened 2018-12-12
Fix list button styling

monty0157:
Fixes styling on main button for a list. Test that creating a List with a Main Button doesn't look like this:

imgo


j-sommer:
LGTM

Just one thing with the styling of the subtitle:
screenshot from 2018-12-12 13-57-14

PR #53, opened 2019-06-19
Establish socket connection lazy

robinschuster:
The socket connection should be established if one of the following event occurs:
- the webchat is opened
- the webchat sends a message (also via webchat.sendMessage api)


arasabbasi:
lgtm!

PR #59, opened 2019-07-15
Expose Webchat's React

robinschuster:
IMPORTANT!!!
It is now mandantory to load the webchat.js before any Webchat Plugins, because the Plugins using this API are dependent on the Webchat being loaded first.

In order to use React's Context and Hooks API correctly, it is mandantory that any part of the React component tree within the Webchat is rendered with the same instance of React. This means that Plugins need a way to access the Webchat's React instance instead of bringing their own. 

This can be done by exposing React under `window.__COGNIGY_WEBCHAT.React` when the Webchat is loaded. Plugins can be bundled with an aliasing to resolve `react` to `window.__COGNIGY_WEBCHAT.React` (this is possible with webpack as well as parcel, i put the webpack configuration in here for builtin plugins).

Another benefit of this is that Plugins are smaller because they do not include their own version of React.


arasabbasi:
It doesnt seem to throw an error, when loaded separately. 

Reproduction Steps:
- copy content of javascript into window.console
- run
Bildschirmfoto von 2019-07-17 10-45-22

arasabbasi:
Probably will be triggered when react method is called.

PR #72, opened 2019-12-09
add input sanitization for outgoing messages

pedily:
When sending a message to from the webchat, the text value will be html-sanitized to avoid XSS attacks.


dshire:
We should add an optional opt-out possibility in case a particular use case requires unsanitized input.

Cesarcs7:
lgtm

PR #75, opened 2019-12-12
6848 - Replace @cognigy/webchat-client

pedily:
The goal of this PR is to get rid of the dependency on @cognigy/webchat-client.

Therefore, the following features moved here from WebchatClient:
- default channel name is set to "webchat-client"
- fetching of endpoint config
- providing of analytics event emitter
- interfaces for WebchatSettings

The API of the Webchat did not change in this PR.
Everything should work exactly as before.


Cesarcs7:
lgtm

PR #81, opened 2020-01-03
(6947) As a Webchat embedder i want to have the possibility to disable autofocus when the webchat is opened

mtnptrsn:
Adding possibility to disable autofocus on input when opening widget.
peek


Cesarcs7:
lgtm

PR #82, opened 2020-01-07
Feature/6905 refactor date picker

kwinto:
Please test with a different language settings of the date-picker node.

We use now a native browser implementation for locale date formatting.
It is discouraged to use this value for operations with a date as we also provide now
an array of dates in ISO format in "data" payload. Backend implemetation comming.

CHANGELOG:
- moment.js lib is removed. We spare 30% of the bunde size (~260K);
- we do not preselect any date for the user;
- disabled state for submit button;


kwinto:
Can one please confirm the difference of date representation with „ko“ or „jp” locale between IE11 and other browsers? 

We should discuss if that might be an issue as we now encourage the use of ISO dates provided in data payload (backend implementation waiting).

PR #91, opened 2020-02-21
Feature/7767 date picker css classes

kwinto:
Please test date picker plugin is functioning and picking the new classes up

### DIFF
--- a/docs/css-customization.md
+++ b/docs/css-customization.md
@@ -548,3 +548,33 @@ This is the button inside every list element, adding some hovering effects or an
   background-color: #cfb00071;   
 }
 ```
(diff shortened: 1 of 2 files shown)


pedily:
Is it possible to style the things that appear within the FlatPicker component?

kwinto:
Yes, all of the Flatpickr elements have static class names. We self use `[data-cognigy-webchat-root] .class_name` to style it though. So one must increase the css specificity of the rule or make sure it is applied after ours.

pedily:
That works, but we still need documentation for that. Please add examples / list selectors to target the specific elements

PR #92, opened 2020-03-03
Implement disableHtmlInputRendering setting

kwinto:
Please test the new `disableHtmlInputRendering` setting. With this set to true, input from the user should be free from the html tags.

Signed-off-by: Dmitrii Ostasevich <d.ostasevich@cognigy.com>


Cesarcs7:
The input being free of HTML tags already works, what this PR fix is the possibility to render a link through <a href=#>link</a>. lgtm

PR #105, opened 2020-05-07
Fix speech input

kwinto:
This PR fixes speech input. 

To test build the speech input plugin and include it after the main bundle.


docs/embedding.md
```diff
@@ -57,7 +57,6 @@ See it in action:
 | getStartedPayload | string | "GET_STARTED" | The payload to send to your Flow when clicking the Get Started Button / when sending the auto message. 
 | getStartedButtonText | string | "GET_STARTED" | The text to display on the Get Started Button / when sending the auto message. 
 | inputPlaceholder | string | "Write a reply" | The placeholder text to display in the input field. 
-| enableSTT | boolean | false | Whether to enable speech input that lets the user speak to the Webchat instead of only typing.
```
pedily, line 60: I think instead of removing this line we should instead add a short notice that the plugin needs to be loaded for this to work.

PR #110, opened 2020-06-09
Bug/9198 fix connection status indicator

pedily:
This PR addresses two issues with the connection status indicator:

# The Indicator stays even if the connection is reestablished
Customers reported that the indicator would randomly show up and stay even if the connection was restablished (e.g. they got messages from the bot while the indicator was there)
I tackled this by setting the "connected" state to "true" in case
- a message arrives from the bot
- a "pong" event is emitted from the socket (this is an internal feature of socket.io, check their docs)

# The Indicator completely blocks mobile views
On mobile views, the indicator blocks any interaction with the webchat, it even blocks the "close" button. This means users on phones have to reload the page in order to close the webchat again.
I tackled this by also rendering a "close" button into the connection status indicator

### DIFF
--- a/src/webchat/store/connection/connection-handler.ts
+++ b/src/webchat/store/connection/connection-handler.ts
@@ -7,7 +7,7 @@ export const registerConnectionHandler = (store: Store, client: SocketClient) =>
         store.dispatch(setConnected(true))
     }
 
-    const handleDicsonnected = () => {
+    const handleDisconnected = () => {
         store.dispatch(setConnected(false));
     }
 
@@ -16,7 +16,7 @@ export const registerConnectionHandler = (store: Store, client: SocketClient) =>
     client.on('socket/pong', handleConnected);
     client.on('output', handleConnected);
 
-    client.on('socket/disconnect', handleDicsonnected);
+    client.on('socket/disconnect', handleDisconnected);
 
     client.on('socket/error', error => {
         const reconnectionLimit = error.type === 'RECONNECTION_LIMIT';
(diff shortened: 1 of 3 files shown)


src/webchat/store/connection/connection-handler.ts
```diff
@@ -3,9 +3,21 @@ import { setConnected, setReconnectionLimit } from "./connection-reducer";
...
 
 export const registerConnectionHandler = (store: Store, client: SocketClient) => {
-    client.on('socket/connect', () => { store.dispatch(setConnected(true)) });
-    client.on('socket/reconnect', () => { store.dispatch(setConnected(true)) });
-    client.on('socket/disconnect', () => { store.dispatch(setConnected(false)) });
+    const handleConnected = () => {
+        store.dispatch(setConnected(true))
+    }
+
+    const handleDicsonnected = () => {
```
dshire, line 10: "handleDicsonnected" unfortunate typo

src/webchat/store/connection/connection-handler.ts
```diff
@@ -3,9 +3,21 @@ import { setConnected, setReconnectionLimit } from "./connection-reducer";
...
+        store.dispatch(setConnected(false));
+    }
+
+    client.on('socket/connect', handleConnected);
+    client.on('socket/reconnect', handleConnected);
+    client.on('socket/pong', handleConnected);
+    client.on('output', handleConnected);
+
+    client.on('socket/disconnect', handleDicsonnected);
```
dshire, line 19: same typo reused

src/webchat/store/connection/connection-handler.ts
mastasky: Please provide more details on how this is unfortunate

dshire: https://en.wikipedia.org/wiki/Freudian_slip

PR #112, opened 2020-07-08
Allow user to output _any_ content

kwinto:
This PR adds user ability (responsibility) to output any HTML content in different message plugins.


pedily:
The html rendering does not work for the "big"-style top element of a list view
Screenshot from 2020-07-09 11-48-52

PR #115, opened 2020-07-30
Fix Date format

sushmi21:
- fixes the date format override bug by removing the custom date formatting ( Fix for first subtask in [https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/9829](url) )
- Adds options to prep for @mastasky  PR


sushmi21:
This bug is fixed in #116

PR #121, opened 2020-09-23
Fix double action on Lists with button and "Default action" url

kwinto:
Test:
- Create a List in SayNode Webchat configuration with a postback button, also set default action url (use v3 as v4 lacks these options now).
- Talk to a bot. 
- Clicking on a button should result in a sent postback and should not open an "Default action" url.


Cesarcs7:
lgtm

PR #122, opened 2020-09-29
Bug/10150 multiline in gallery card descriptions not converted correctly

Cesarcs7:
Previously when using Multiline in gallery card descriptions subtitle (Default tab, Gallery cards from Carousel)
The rendered card in the webchat will not display the line breaks,

This PR adds a flag "ignoreLineBreak" and also puts the line break as a default behaviour.

Setting the flag to true will make that the cards fall back to the old behaviour, this way our customers will have the choice to let things like the used to have them before.

To test:
- Create a simple Gallery Carousel in a Say node and add multiline text in the subtitle field, also add spaces in some lines.
- Talk to this from the webchat without activating the flag and also activating the flag. You should see the difference.

### DIFF
--- a/src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
+++ b/src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
@@ -29,9 +29,9 @@ export interface IMessengerGenericTemplateState {
 
 export const getMessengerGenericTemplate = ({
     React,
-    styled
+    styled,
 }: MessagePluginFactoryProps) => {
-    const MessengerSubtitle = getMessengerSubtitle({ React, styled });
+    const MessengerSubtitle = getMessengerSubtitle({ React, styled});
     const MessengerTitle = getMessengerTitle({ React, styled });
     const MessengerFrame = getMessengerFrame({ React, styled });
     const MessengerContent = getMessengerContent({ React, styled });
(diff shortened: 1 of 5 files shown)


src/plugins/messenger/MessengerPreview/components/MessengerSubtitle.tsx
```diff
@@ -1,11 +1,13 @@
...
 
 export const getMessengerSubtitle = ({ React, styled }: MessagePluginFactoryProps) => {
-    const MessengerSubtitle = styled.p({
+    const MessengerSubtitle = styled.p<{ config: IWebchatConfig }>(props => ({
         margin: 0,
         color: 'hsla(0, 0%, 0%, .54)',
-        fontSize: 13
-    });
+        fontSize: 13,
+        whiteSpace: props.config.settings.ignoreLineBreaks ? "initial" : "pre-wrap" ,
```
kwinto, line 9: We should use "pre-line" instead of "pre-wrap". "pre-wrap" also changes the way sequence white spaces are treated which seems to be an unwanted change for this bugfix.

Cesarcs7: fixed

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -29,9 +29,9 @@ export interface IMessengerGenericTemplateState {
 
 export const getMessengerGenericTemplate = ({
     React,
-    styled
+    styled,
 }: MessagePluginFactoryProps) => {
-    const MessengerSubtitle = getMessengerSubtitle({ React, styled });
+    const MessengerSubtitle = getMessengerSubtitle({ React, styled});
```
dshire, line 34: important change :smile:

dshire (approved):
lgtm

kwinto (approved):

PR #129, opened 2020-10-23
Change style of Button Template

mastasky:
Colour button template top in theme primary color
Line breaks in Button Template top

### DIFF
--- a/src/plugins/messenger/MessengerPreview/components/MessengerButtonTemplate/MessengerButtonTemplate.tsx
+++ b/src/plugins/messenger/MessengerPreview/components/MessengerButtonTemplate/MessengerButtonTemplate.tsx
@@ -3,8 +3,7 @@ import { IWithFBMActionEventHandler } from '../../MessengerPreview.interface';
 import { getDivider } from '../Divider';
 import { MessagePluginFactoryProps } from '../../../../../common/interfaces/message-plugin';
 import { getMessengerButton } from '../MessengerButton/MessengerButton';
-import { getMessengerFrame } from '../MessengerFrame';
-import { getMessengerButtonTop } from '../MessengerButtonTop';
+import { getMessengerButtonHeader } from '../MessengerButtonHeader';
 
 interface IMessengerButtonTemplateProps extends IWithFBMActionEventHandler {
     payload: IFBMButtonTemplatePayload;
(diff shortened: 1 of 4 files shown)


src/plugins/messenger/MessengerPreview/components/MessengerButtonTop.tsx
```diff
@@ -0,0 +1,14 @@
+import { MessagePluginFactoryProps } from '../../../../common/interfaces/message-plugin';
+
+export const getMessengerButtonTop = ({ React, styled }: MessagePluginFactoryProps) => {
+    const MessengerFrame = styled.div(({ theme }) => ({
+        width: 250,
+        borderRadius: theme.unitSize * 2,
+        overflow: 'hidden',
+        background: theme.primaryGradient,
+        color: theme.primaryContrastColor,
```
dshire, line 9: Looks nice with the theming, but why not add it everywhere then (lists, galleries)? Also, we should probably make this an optional config setting as it changes previous behavior

dshire (changes requested):

src/plugins/messenger/MessengerPreview/components/MessengerButtonTop.tsx
mastasky: Lists and Galleries are different Buttons are more like QRs. I consider it a bug before. Let's keep only here.

mastasky:
@pedily  - please make sure that linebreaks can be turned off with setting

dshire (approved):

src/plugins/messenger/MessengerPreview/components/MessengerButtonTemplate/MessengerButtonTemplate.tsx
```diff
@@ -13,12 +13,13 @@ export const getMessengerButtonTemplate = ({
...
 }: MessagePluginFactoryProps) => {
-    const MessengerFrame = getMessengerFrame({ React, styled });
+    const MessengerButtonHeader = getMessengerButtonHeader({ React, styled });
     const MessengerButton = getMessengerButton({ React, styled });
     const Divider = getDivider({ React, styled });
 
     const Text = styled.div(({ theme }) => ({
-        padding: theme.unitSize
+        padding: `${theme.unitSize * 2}px ${theme.unitSize * 3}px`,
+        whiteSpace: "pre-line"
```
mastasky, line 22: @pedily Please make this optional via config

kwinto (approved):
lgtm

PR #130, opened 2020-11-02
Feature/11288 engagement message

pedily:
This PR adds the concept of an "engagement message" which will imitate an incoming message from the bot after a delay.
The feature can be used by configuring `engagementMessageText` in the webchat settings.
The delay can be set in milliseconds by changing `engagementMessageDelay` (defaults to 5000).

When the delay runs out, the engagement message will be triggered under the following conditions:
- the `engagementMessageText` setting is not empty
- ~~the message history is empty~~
  EDIT: removed this
- the endpoint is active
- the webchat has not been opened yet

The engagement message will not show up in the chat history ui unless set so by showEngagementMessageInHistory

This PR also adds some tests which can be run by executing `npm test` after running `npm ci`.

The test setup required some changes to the webpack dev server:
The port of the webserver is now running at port 8787 (it had to be something that we are sure is not taken yet so we can hardcode the port in tests).

You can now run `npm run cypress:open` to open the cypress tests. They will use the webpack dev server build and live-reload on saving webchat or test code.

The embedding code in `dist/index.html` will now only run under the condition that we are not executing tests right now


pedily:
if autoSendMessage is used in combination with engagementMessage, the autoMessage should be sent if the history contains nothing but engagement messages

mastasky (approved):
lgtm

docs/embedding.md
```diff
@@ -72,6 +72,8 @@ See it in action:
 | enableUnreadMessagePreview | boolean | false | | | If true, the webchat shows a message bubble with the latest retrieved bot message.
 | enableUnreadMessageSound | boolean | false | x | | If true, plays a notification sound for each incoming unread message  
 | enableUnreadMessageTitleIndicator | boolean | false | | If true, will indicate the amount of unread messages in the page title every 1000ms
+| engagementMessageText | string | "" |  |  | If this is set to a non-empty string, there will be an automatic "fake" engagement message being triggered after a delay
+| engagementMessageDelay | number | 5000 | | | The timeout after which the engagement message should be triggered
```
kwinto, line 76: Should mention the unit of measure

kwinto:
`engagementMessageText: true` should overrule `enableUnreadMessageBadge: false`
otherwise it is not possible to have an "engagement message" and NOT have previews for regular messages at the same time

kwinto:
`enableUnreadMessageTitleIndicator: true` with `engagementMessageText: "foobar"` make sense only with combination of `showEngagementMessagesInChat: true`. Otherwise the user gets a "(1)" badge and sees an empty chat on open which is confusing

kwinto:
>When the delay runs out, the engagement message will be triggered under the following conditions:
>
>the engagementMessageText setting is not empty
>**the message history is empty**
>the endpoint is active
>the webchat has not been opened yet

how is this criteria (in bold) mean to be met? there *might* be a chat history, but it is loaded only after the webchat is open

kwinto:
>sh: http-server: command not found
on `npm test`, possibly missing dependency

kwinto:
`startBehavior.spec.ts`

2/3 test are failing

sushmi21:
Two tests are failing for me in `startBehavior.spec.ts` as Dmitrii mentioned.

```
 1) Start Behavior
       should automatically send a preconfigured "get started" message:
     AssertionError: Timed out retrying: Expected to find content: 'get started' but never did.
      at Context.eval (http://localhost:8787/__cypress/tests?p=cypress/integration/startBehavior.spec.ts:111:47)

  2) Start Behavior
       should automatically send a "get started message" even if the history contains an engagement message:
     AssertionError: Timed out retrying: Expected to find content: 'get started' but never did.
      at Context.eval (http://localhost:8787/__cypress/tests?p=cypress/integration/startBehavior.spec.ts:142:36)
```

docs/embedding.md
pedily: fixed

pedily:
The two failing tests are failing because we do not have an active connection and changed the injection behavoir to trigger after the connection. I disabled them here until we improved our testing methods so we can fake a connection

sushmi21:
tests look good to me now!

pedily:
regarding "history is empty" condition: i updated the description. the idea of the engagement message is to get the user to open the webchat when there hasn't been any activity after page load. so it's okay to show them even if we had a session before

pedily:
regarding "implicitly enabling / disabling settings depending on another setting": 
i would not do that. sure, you have to configure a set of settings in order to get it running properly, but you lose the power to configure it in another way.

PR #137, opened 2020-12-01
(re-)store only relevant information from storage

kwinto:
This PR changes what we store and restore for the "persistence history" mechanism to the local (-session) storage
- it is only the "messages" now.

Also fixes storing duplicate data with an empty keys (["","",""]) in storage.

To test: 
initialize the webchat with a fixed sessionId string and test it behaves normally, also with these endpoint settings:
                    `getStartedText: "INJECT",
                    startBehavior: "injection"`
and these:
                    `disablePersistentHistory: true,
                    engagementMessageText: "Hi!",
                    engagementMessageDelay: 500`

### DIFF
--- a/src/webchat/store/options/options-middleware.ts
+++ b/src/webchat/store/options/options-middleware.ts
@@ -11,6 +11,7 @@ export const optionsMiddleware: Middleware<{}, StoreState> = store => next => (a
     const key = getOptionsKey(store.getState().options);
     const { active } = store.getState().config; // Actual settings are loaded
     const { disableLocalStorage, disablePersistentHistory, useSessionStorage } = store.getState().config.settings;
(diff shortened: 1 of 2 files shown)


dshire (changes requested):
You also need to keep the ui.open state, else the chat closes on a multi-page website implementation when switching the page

kwinto:
> You also need to keep the ui.open state, else the chat closes on a multi-page website implementation when switching the page

You mean this is reproducible on this branch? The partial saved state is being merged here, therefore it should actually be independent.

dshire (approved):

dshire:
as discussed the "open" state is not stored in localstorage any longer and only the messages are. this is working fine with these changes so I approved.

sushmi21 (approved):

PR #139, opened 2020-12-07
Automated tests on PR

kwinto:
This PR adds a "github workflow" to run our test suite after creating a PR. Subsequent pushes to the branch of PR will retrigger test run.

Also fixes currently failing test (minor issue).

After this get merged, we need to setup branch protection on `master` to prohibit non-admin merge without successful test run.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -6,5 +6,5 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
-      - run: npm install
+      - run: npm ci
       - run: npm test
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,0 +1,10 @@
+name: Run Cypress
+on: [pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,10 @@
+name: Run Cypress
+on: [pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm install
```
pedily, line 9: shouldn't this be `npm ci`?

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,10 @@
+name: Run Cypress
+on: [pull_request]
```
pedily, line 2: will this also reset / trigger again if new content is pushed to a pr?

.github/workflows/cypress.yml
kwinto: fixed

.github/workflows/cypress.yml
kwinto: see pr description

PR #145, opened 2021-01-18
Feature/force polling

pedily:
This PR upgrades the socket-client to version 4.5.0 which supports the new `disableWebsockets` flag.

Please test that
| forceWebsockets | disableWebsockets | expected result |
| - | - | - |
| `false` | `undefined` or `false` | first connection via http, then upgrade to websockets |
| `true` | `undefined` or `false` | direct websocket connection without upgrade |
| `true` | `true` | _same outcome as row above_ |
| `false` | `true` | connection via http polling without trying to upgrade to websockets (one http request per user message etc) |

### DIFF
--- a/docs/embedding.md
+++ b/docs/embedding.md
@@ -1,7 +1,9 @@
 # Embedding the Cognigy Webchat
 
 ## Basic Implementation
+
 To integrate the Webchat into a Website, you need to
+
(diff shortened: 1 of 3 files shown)


kwinto:
both options set to true result in socket connection, other seem fine

kwinto (approved):

dshire (approved):

PR #147, opened 2021-01-19
add "debounced" typing indicator

pedily:
This PR moves the typing indicator into its own component and adds a "one-sided debounce" to it.
When the indicator is set to "on", it immediately appears, but when it's turned "off" it will stay a bit longer.
This way, multiple incoming messages don't cause the indicator to look "flickering".

Unfortunately, I had to add a timeout to the explicit "scrollToBottom" after sending a message so the scroll position would behave correctly. I'm open to ideas to change that!


dshire (changes requested):
I definitely get your approach here, and it looks nice to have the typing indicator stay in one fluid motion, especially with multiple messages coming in.
However, the behavior when a message appears doesn't look nice anymore, as the timeout makes it that the typing indicator stays even when the message already came in. It should disappear with the incoming message, ideally it should "become" the incoming message. 
Let's have a look at the old and new indicator with a single incoming message:
previous:

https://user-images.githubusercontent.com/30043753/105338167-86a72b00-5bdb-11eb-87ed-f34903f86910.mp4

changed:

https://user-images.githubusercontent.com/30043753/105338261-a6d6ea00-5bdb-11eb-9e2b-d88e2ff2bed7.mp4

### My points:
1. the avatar should be added to the the typing indicator
2. the typing indicator should not stay longer than the message "it is typing", ideally it should "become" the new message (see top video). If we don't want this, then the indicator should not interfere with scroll position OR we need smooth scrolling of new incoming messages. But using the previous behavior as a starting point, the "Dots" become the message text.
3. The "Powered by Cognigy.AI" seems to disappear occasionally when the typing indicator appears (see bottom video). This seems to be related to scrolling, the "powered by" gets pushed to the bottom and out of view

pedily:
@dshire 
the "indicator becoming the new message" was only nice if the incoming message had the same shape, color, and size as the "indicator bubble". If you expand your example with a multi-line text, a gallery, a text with quick replies, or any other type of custom plugin message, the effect will not work anymore.

Please extend your example to a bot that sends 10 messages of different size and shape.
As soon as you are "scrolled to the bottom" and receive a handful of not-single-line-of-text messages, it looks ugly.

> the avatar should be added to the the typing indicator

We don't know the correct avatar before the message arrives. It might be coming from the bot or the agent.

> the typing indicator should not stay longer than the message "it is typing", ideally it should "become" the new message (see top video). If we don't want this, then **the indicator should not interfere with scroll position** OR we need smooth scrolling of new incoming messages. But using the previous behavior as a starting point, the "Dots" become the message text.

I think we should try an approach where the indicator does not take up additional space in the history but is shown or hidden in a reserved space after the last message so the layout won't "jump" when it appears or disappears.
Afterward, we can then decide whether debounding or directly toggling the indicator is nide.

> The "Powered by Cognigy.AI" seems to disappear occasionally when the typing indicator appears (see bottom video). This seems to be related to scrolling, the "powered by" gets pushed to the bottom and out of view

this is the issue I was trying to address with the "setTimeout" in WebchatUI, but it seems to be inconsistent.
However, if we go for the "typing indicator does not take up space in history" approach, this will no longer be an issue.

dshire:
yeah, I tested it with mutiple rich and multi-line messages coming in first, and when I noticed the issues I just created a simple example since this is already valid there.

I think going with a reserved space sounds fine, should just be less than it currently takes. I think the indicator should definitely not stay longer than any message is coming in.

kwinto:
>if we go for the "typing indicator does not take up space in history" approach, this will no longer be an issue
this has a trades-off on valuable vertical space. On mobile devices with keyboard on the screen user will hardly see any context.

should we consider doing "debounce" but on messages output themselves and not typing indicator?

kwinto:
Also I can only reproduce this "flickering" effect on master branch state in Firefox, but not in Chrome or Safari.

pedily:
@kwinto @dshire i implemented a "reserved space" variant and minimized the used space.

Here's a screen recording from my phone

dshire (approved):
Working as expected. I still think there is room to improve how this looks and feels, but most importantly this fixes the flickering.

kwinto:
Actually not bad visually! The only thing, it stays awhile after the message arrives. This might be a matter of habit, but it causes in me a strong cognitive dissonance - "how could someone be typing after he has already sent the message".

Is it possible to fix this? I thought reserving space alone should be sufficient to avoid "flickering" effect.

pedily:
@kwinto you would still see the typing indicator "jittering around" because it gets toggled so much.

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -289,11 +290,17 @@ export class WebchatUI extends React.PureComponent<React.HTMLProps<HTMLDivElemen
...
+
         if (this.history.current) {
-            this.history.current.scrollToBottom();
+            /**
+             * this timeout is necessary to avoid
+             * an immediate "typing indicator" to not be scrolled correctly
+             */
+            setTimeout(() => {
+                this.history.current.scrollToBottom();
+            }, 150);
```
kwinto, line 302: Is this still necessary with a latest changes?

pedily: fixed

kwinto (approved):

PR #149, opened 2021-02-04
Fix width of avatar in chat

sushmi21:
The width of the avatar was 26px and height was 24px, making the avatar look off. This PR fixes the width of the avatar in the chat, to make it a circle with equal width and height (24px). 

To test: 
- Test the avatar of both the sender and the receiver

### DIFF
--- a/src/webchat-embed/embedded-webchat-styles.css
+++ b/src/webchat-embed/embedded-webchat-styles.css
@@ -15,10 +15,6 @@
     z-index: 2;
 }
 
-[data-cognigy-webchat-root] [data-cognigy-webchat].webchat .webchat-avatar { 
-	max-width: 24px;
-}
-
 [data-cognigy-webchat-root] button[data-cognigy-webchat-toggle] {
     position: fixed;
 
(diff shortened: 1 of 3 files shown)


pedily:
This change should be done in `MessageRow.tsx` by changing the flex-basis to 24 instead of overriding it at another place.

src/webchat-embed/embedded-webchat-styles.css
```diff
@@ -15,6 +15,9 @@
     z-index: 2;
 }
 
+[data-cognigy-webchat-root] [data-cognigy-webchat].webchat .webchat-avatar {
```
pedily, line 18: This change should be done in `webchat-ui/components/presentational/MessageRow.tsx:57` by changing `flexBasis` to `24` instead of overriding it here.

pedily (changes requested):

src/webchat-embed/embedded-webchat-styles.css
sushmi21: fixed

pedily (approved):

Cesarcs7 (approved):

PR #150, opened 2021-02-16
remove race condition causing "auto-send message" not to work

pedily:
This PR fixes an issue that causes "auto-send messages" to not trigger properly.
The problem can occur when using "auto-send message" together with `webchat.open()` in the embedding code, causing a race condition.

The "before state" looked like this:
`initWebchat()` will trigger `load config`
`open webchat` will trigger `connect`
`connected` will trigger `handle injection message`
`handle injection message` checks if "injection message" is configured in the config and, if configured, will send the injection message

The problem:
if `connected` happens before `config loaded`, the `handle injection message` was executed without access to the endpoint configuration for "injection message".

The solution:
The "injection message" was moved to its own reducer/middleware.
The reducer registers whether certain things already happened once:
- webchat was opened
- connection was established
- config was loaded
- injection message was handled
It will listen to the `SET_OPEN`, `SET_CONNECTED`, and `SET_CONFIG` actions used by the other reducers and store whether they already happened.
Additionally, it will also register whether the "injection message" happened via a new action.

The middleware will listen to `SET_OPEN`, `SET_CONNECTED`, and `SET_CONFIG`events, handle them as usual, then check whether the "injection message" was not already handled and check whether it's ready.
If it's ready and not handled, it should be triggered now.

This fixes the "order of events" issue.

# How to Test
- configure a webchat endpoint with an "injection message"
- run the webchat from this branch towards that endpoint
- configure your embedding code to open the webchat immediately
```javascript
initWebchat('').then(webchat => {
  webchat.open();
});
```
- refresh the browser tab multiple times
  - without this fix, the injection message may not be sent
  - with this fix, the injection message should always be sent

### DIFF
--- a/src/webchat/store/autoinject/autoinject-middleware.ts
+++ b/src/webchat/store/autoinject/autoinject-middleware.ts
@@ -1,9 +1,6 @@
-import { SocketClient } from "@cognigy/socket-client";
 import { Middleware } from "redux";
 import { StoreState } from "../store";
 import { autoInjectHandled, TAutoInjectAction, triggerAutoInject } from './autoinject-reducer';
-import { getAvatarForMessage, sendMessage } from '../messages/message-middleware';
-import { addMessage } from "../messages/message-reducer";
 import { Webchat } from "../../components/Webchat";
 
 export const createAutoInjectMiddleware = (webchat: Webchat): Middleware<unknown, StoreState> => api => next => (action: TAutoInjectAction) => {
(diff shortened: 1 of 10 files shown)


src/webchat/store/autoinject/autoinject-middleware.ts
```diff
@@ -0,0 +1,72 @@
...
+import { getAvatarForMessage, sendMessage } from '../messages/message-middleware';
+import { addMessage } from "../messages/message-reducer";
+import { Webchat } from "../../components/Webchat";
+
+export const createAutoInjectMiddleware = (webchat: Webchat): Middleware<unknown, StoreState> => api => next => (action: TAutoInjectAction) => {
+    switch (action.type) {
+        case 'SET_CONFIG':
+        case 'SET_CONNECTED':
+        case 'SET_OPEN': {
+            next(action);
```
kwinto, line 14: Don't you want to save the result and return it afterwards on :20, :23, :27 (also on :71)? So that a future code may rely on a convention of "dispatch() can return a value"?

pedily: The idea here is that we are registering a "side effect" to these three events. The checks we want to do rely on the events to be fully processed, therefore we immediately call `next(action)` and use `return` later-on in order to avoid the action from being passed on twice.

After calling `next(action)`, using `api.getState()` provides us with the "up-to-date" state (the action has been processed by all middlewares and reducers), safe for further checks.

If the conditions are met after the action was processed, we dispatch an additional action as a "side-effect".

kwinto (approved):

sushmi21 (approved):

PR #156, opened 2021-03-05
Ignore "engagement message" in "get started button" rule

pedily:
This PR fixes an issue that would render the "get started button" useless if you used it with an engagement message since the message would be in the history and would cause the "get started button" to not render.

To solve this, I added an exception for the "get started button" rule to also show if there is only an engagement message in the chat.

Since there can only be one engagement message, we can implement this check cheaply.

To test this, apply the following settings to your webchat:

```javascript
{
  settings: {
    enableUnreadMessagePreview: true,
    showEngagementMessagesInChat: true,
    getStartedText: "Text",
    getStartedButtonText: "Button Text",
    getStartedPayload: "Payload",
    startBehavior: 'button',
    engagementMessageText: "Engagement Message",
    engagementMessageDelay: 1000
  }
}
```

Without the patch, the "get started button" will not be shown when the webchat was opened after the "engagement message" triggered. With the patch applied, it should still be shown.


kwinto (approved):
lgtm, but it would be nice to have a test case for this.

PR #157, opened 2021-03-08
Fix landmark role and chat header accessibility

sushmi21:
This PR fixes following accessibility issues in webchat.
 - Add heading role to webchat heading
 - Fix aria-label for 'Powered by Cognigy' text
 - Add ids to chat header , chat window and chat button
 - Add focus style for get started button


pedily:
axe is complaining that the "powered by cognigy" banner is too low-contrast (which is by design, I guess, should be subtle).

Can we tell the screenreader to treat this as a symbol instead of text so it won't complain?

sushmi21:
> axe is complaining that the "powered by cognigy" banner is too low-contrast (which is by design, I guess, should be subtle).
> 
> Can we tell the screenreader to treat this as a symbol instead of text so it won't complain?

Powered by texts need not have high contrasts. They need not be accessible.

pedily (approved):

PR #159, opened 2021-03-23
Hide a message from the history if empty GetStartedText

kwinto:
This PR fixes a regression with an empty `getStartedText` option resulted in `getStartedPayload` being posted in a chat. in that case user expects that an empty option will hide the message bubble entirely (as it was before). This is now fixed.

To test use these options for Webchat:
```
getStartedText: "",
getStartedPayload: "foo",
startBehaviour: "injection",
```

After opening the Webchat you should get the reply from the bot
but should not see `foo` in the message history


dshire (approved):
works as expected

sushmi21 (approved):

PR #161, opened 2021-03-26
Prevent injection if session restored with messages

kwinto:
This PR makes sure the injection message is not being send if the Webchat session restored with some messages.

To test use this settings:
```
        initWebchat('https://endpoint-dev.cognigy.ai/f7d316bceb80ea4afeba518fccf563153e6f8d174d292bbf73f8d15bba025a5e', {
            userId: "user1",
            sessionId: "session1",
            settings: {
                getStartedText: "start!",
                startBehavior: "injection",
            }
        }).then(webchat => {
            window.cognigyWebchat = webchat;
            webchat.open();
        });
```

if you reload the page after the first load the message "start!" should not be sent

### DIFF
--- a/cypress/integration/startBehavior.spec.ts
+++ b/cypress/integration/startBehavior.spec.ts
@@ -16,7 +16,7 @@ describe('Start Behavior', () => {
             .contains('get started').should('be.visible');
     });
 
-    it('should not send a "get started message" if the history contains messages', () => {
+    xit('should not send a "get started message" if the history contains messages', () => {
         cy
             .visitBuild()
             .initMockWebchat({
@@ -32,7 +32,7 @@ describe('Start Behavior', () => {
             .contains('get started').should('not.exist');
     });
 
-    it('should not send a "get started message" if the getStartedText is empty', () => {
+    xit('should not send a "get started message" if the getStartedText is empty', () => {
         cy
             .visitBuild()
             .initMockWebchat({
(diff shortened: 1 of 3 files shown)


pedily:
I think the approach you chose is not safe to use.

If you dispatch `SET_OPTIONS`
- ...it's processed by other middlewares
- it's processed by the `autoInjectMiddleware`
  - we register a side-effect, checking whether `autoInject.isSessionRestoredOnce` is `true` 
  - it's not `true` yet, so we abort
- it's processed by the optionsMiddleware
  - as a side-effect, `RESET_STATE` may be called with the content from the `localStorage` (asynchronous)
- it's processed by the reducers, causing the `options` reducer to be updated and the `autoInject` reducer setting `isSessionRestoredOnce` to `true`

Your implementation will not cause the `SET_OPTIONS` action to trigger the "auto inject" path without other side effects.

In order to fix this, I would suggest to create a new `SESSION_RESTORED` action, which is dispatched after the session was either restored or not (behind `options-middleware.ts:31).

Basically, like this:
```
case 'SET_OPTIONS': {
  if(storage) {
    ...
  }
  const result = next(action);
  dispatch(sessionRestored());
  return result;
}
```

This would make sure the `SET_OPTIONS` action is processed by the `autoInject` reducer first and then trigger a new pipeline run with the new action afterward.
That new action will be processed by the `autoInjectMiddleware` and the state for `autoInject.isSessionRestoredOnce` is already `true`.
As the things build up on top of each other, we do have a guaranteed, race-condition-free chain of side effects.

pedily:
Update from my side: my concerns were not problematic at all, since the `injectMiddleware` already takes care of the event to be processed before checking for the new state.

pedily (approved):

PR #165, opened 2021-03-31
Scroll card with focused button into view

sushmi21:
When you have gallery cards with buttons, you will be able to focus the buttons in the cards using tab. Sometimes, when the button in the card receives focus, the card will not be scrolled to the middle (Example: When you have three cards with at least one button each, when a button in card 2 receives focus, the card will not be scrolled to middle). Hence, you will be able to see only half of the focused button and the card in which it is present. 
With this fix, the button inside the card that is focused will be scrolled into view. Additionally, the button groups are assigned aria-labelledby and aria-describedby attributed

To test:
- Create gallery with multiple cards and add at at least one button to each card 
- Navigate using keyboard and see if the focused button is fully visible


kwinto (approved):
ie11 ✔️

Cesarcs7 (approved):

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -105,27 +112,37 @@ export const getMessengerGenericTemplate = ({
                     )
             ) : null;
 
+            const carousalRootId = `webchat-carousel-template-root-${uuid.v4()}`;
```
dshire, line 115: typo *carousel

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -105,27 +112,37 @@ export const getMessengerGenericTemplate = ({
...
+            const carouselTitle = `webchat-carousel-template-title-${uuid.v4()}`;
+            const carouselSubtitle = `webchat-carousel-template-subtitle-${uuid.v4()}`;
+            // add aria-describedby only if subtitle is present
+            const carousalAriaDescribedBy = document.getElementById(carouselSubtitle) ? carouselSubtitle : undefined;
```
dshire, line 119: also typo *carousel

dshire (changes requested):
working as expected, there's just some typos you should fix before we merge

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
sushmi21: fixed

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
sushmi21: fixed

dshire (approved):

PR #166, opened 2021-03-31
Trap focus within chat window

sushmi21:
Current behavior: During Tab/ Shift + Tab navigation, focus can be moved to elements outside the chat window. 
Expected behavior: When enableFocusTrap is set to true, focus should never leave the chat window. By default, enableFocusTrap is false, so focus can leave the chat window.

To test: 
- Add webchat setting `enableFocusTrap: true`
- Navigate using keyboard and see is the focus does not leave the chat. The focus order should also be logical during both Tab and Shift + Tab navigation
- Also test by setting enableFocusTrap to false. There should not be any change in the behavior


src/common/interfaces/webchat-config.ts
```diff
@@ -19,6 +19,7 @@ export interface IWebchatSettings {
     enableConnectionStatusIndicator: boolean;
     showEngagementMessagesInChat: boolean;
     enableFileUpload: boolean;
+    enableFocusTrap: boolean;
```
pedily, line 22: you will also need to add an entry to `docs/embedding.md` explaining what this flag does

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -314,6 +314,47 @@ export class WebchatUI extends React.PureComponent<React.HTMLProps<HTMLDivElemen
             />
         );
     }
+    
+    handleKeydown = (event) => {
+        const closeButtonInHeader = document.getElementById("webchatHeaderCloseButton");
```
pedily, line 319: Relying on document.getElementById seems kind of "hacky" as we might accidently collide with other elements on the page.

Would it be possible to use `refs` here? Then we would explicitly assign dom element instances within the webchat.

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -368,6 +409,7 @@ export class WebchatUI extends React.PureComponent<React.HTMLProps<HTMLDivElemen
 							className="webchat-root"
 							aria-labelledby="webchatHeaderTitle"
 							role="region"
+							onKeyDown={enableFocusTrap && open ? this.handleKeydown : () => {}}
```
pedily, line 412: I'd move the `open` check to `handleKeyDown` and always trigger it.

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -314,6 +314,47 @@ export class WebchatUI extends React.PureComponent<React.HTMLProps<HTMLDivElemen
...
+                } else if (getStartedButton) {
+                    getStartedButton.focus();
+                } else if (webchatInputButtonMenu) {
+                    webchatInputButtonMenu.focus();
+                } else {
+                    chatToggleButton?.focus()
+                }
+            }
+        }
+        if(event.target === chatToggleButton) {
```
pedily, line 340: These cases deserve a line of comment or two. What is special about pressing (shift)+tab on these two elements?

pedily (changes requested):

src/webchat-ui/components/WebchatUI.tsx
sushmi21: Added comments

src/webchat-ui/components/WebchatUI.tsx
sushmi21: Fixed

src/common/interfaces/webchat-config.ts
sushmi21: Fixed

kwinto (approved):
Works as expected

src/webchat-ui/components/WebchatUI.tsx
sushmi21: Replaced ids with refs wherever possible.

dshire (approved):
working as expected

PR #168, opened 2021-04-03
Fix missing "get started button" messages

pedily:
This is a followup PR to https://github.com/Cognigy/WebchatWidget/pull/159.

The PR solved the issue by removing the "implicit fallback" to the "get started payload", but it also changed the behavior to "not adding a message at all" while it should be "add a message with empty text content".

Messages without text content will not be rendered by default and result in the same end-user experience, while developers can still rely on tracking the empty message in the chat history (the get started button itself e.g. can vanish immediately instead of staying there until a response appeared)

To test this:
- set up a flow which sends a text message after 3 seconds
- set up an endpoint with a "get started button" with no text but a payload
- point your webchat to the endpoint, don't use persistent history (so the get started button will show)
- click the "get started" button
  - it should immediately disappear on click, not "when the answer arrived"

### DIFF
--- a/src/plugins/get-started-button-input/GetStartedInput.tsx
+++ b/src/plugins/get-started-button-input/GetStartedInput.tsx
@@ -16,10 +16,10 @@ const GetStartedButton = styled(Button)(({ theme }) => ({
 export default ({ onSendMessage, config }: InputComponentProps) => (
     <Toolbar>
         <GetStartedButton
-            onClick={() => onSendMessage(config.settings.getStartedPayload, null, { label: config.settings.getStartedText })}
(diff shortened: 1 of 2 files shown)


kwinto:
With these settings does not work as described. 
Button stays until a message arrives

```           
            settings: {
                getStartedText: "",
                startBehavior: "button",
            }
```

pedily:
@kwinto the configuration you provided is incomplete, you will partially inherit from your endpoint.

You will at least need "button text", "payload" and "behavior"

```
getStartedText: "",
getStartedButtonText: "button label",
getStartedPayload: "get started payload",
startBehavior: "button",
```

kwinto (approved):
lgtm

PR #169, opened 2021-04-06
"Update settings" api

kwinto:
This PR adds the possibilities to update the settings after the WebchatWidget Initialisation.
e.g. 
```
        webchat.updateSettings({
            enableUnreadMessagePreview: false,
        });
```

enableUnreadMessagePreview render logic was update to allow hidding previews on setting flip.


src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -202,6 +202,10 @@ export class WebchatUI extends React.PureComponent<React.HTMLProps<HTMLDivElemen
             }
         }
 
+        if (prevProps.config.settings.enableUnreadMessagePreview && !this.props.config.settings.enableUnreadMessagePreview) {
```
pedily, line 205: This catches the "from on to off" switch, but not the other way round.

For "from off to on", we have to run the logic from 182 to 197 so the `lastUnseenMessageText` gets updated correctly when re-enabling the feature.

kwinto: fixed

PR #170, opened 2021-04-07
Add role group for button groups

sushmi21:
This PR adds a 'group' aria role for a group of buttons and to each card item in a gallery of cards. A label and description will also be added, when one is present.
Success criteria:
- When a button inside the button group is focused, the screen reader should announce it.
- When a label or description is present, it should also be announced by the screen reader.
- When a button inside a gallery card gets focused, there should be a clear differentiation on which card the button belongs to.

To test: 
- Create Say node and configure default tab to include text with buttons. And more than one button.
- Create Say node and configure default tab to include quick replies. Add more than one button.
- Create Say node and configure default tab to include a gallery. Add more than one card with at least one button each.
All these button groups must pass the success criteria. Also, if you know other ways of creating button groups and if they happen to be not accessible, please let me know.


Cesarcs7 (approved):

kwinto:
What is the reasoning behind grouping only if there is more than one element? As we use a group to label that a button(s) is part of, e.g. gallery card, would it make sense to always provide this accessibility hint?

kwinto (changes requested):
Please answer the question above. Otherwise lgtm

sushmi21:
> What is the reasoning behind grouping only if there is more than one element? As we use a group to label that a button(s) is part of, e.g. gallery card, would it make sense to always provide this accessibility hint?

A group role should be used only if there is more than one item within the group. When it comes to a gallery card, the whole card is a group. So, even if there is no button within the card, the card itself will have the group role. Does this answer your question?

kwinto:
Thanks for the explanation!

kwinto (approved):

PR #173, opened 2021-04-13
Auto focus element in latest incoming message

sushmi21:
This PR adds a feature flag 'enableAutoFocus', which when enabled will move the focus to the first interactive element within the latest incoming message. The focus will not be moved when the currently focused element is not within the chat history. This feature will be helpful to maintain the focus order for keyboard users. 

To test:
- Set enableAutoFocus to true in webchat settings
- Use a flow that includes a lot of interactive elements (List with links/ buttons, audio, video, gallery items with links/ buttons, quick replies with buttons)
- Interact with the bot. Only when an element within the chat log is currently focused, the focus should automatically move to the incoming message's first interactive element. That is, if the focus is currently on message input, chat toggle button, close chat button or somewhere outside the chat log/ chat window, the focus will not be moved automatically. 
- On clicking enter, the focused item should be activated.
- When enableAutoFocus is false, the focus should not be moved.


Cesarcs7:
When opening the webchat for the first time, if I already have messages there. the autofocus will not have an effect. If I close and open the webchat again, then it takes place. What is the desired behavior?

sushmi21:
> When opening the webchat for the first time, if I already have messages there. the autofocus will not have an effect. If I close and open the webchat again, then it takes place. What is the desired behavior?

Was the text input focused at the time? If so, the focus will not move

sushmi21:
> > When opening the webchat for the first time, if I already have messages there. the autofocus will not have an effect. If I close and open the webchat again, then it takes place. What is the desired behavior?
> 
> Was the text input focused at the time? If so, the focus will not move

The issue here was, the focus was on chat toggle button when you open the chat. So, according to previous logic, the focus automatically moved to the latest element within the chat log (as the current focus was not on input field). I have made a change here. Now, the focus will move automatically, only if an element within the chat log has the current focus.

Cesarcs7 (approved):
lgtm

kwinto:
We have recently introduced the api to change webchat settings "on-the-fly" by calling webchat.updateSettings() method. You may want to remove `config` from the useEffect dependencies, because calling this method leads to the focus being shifted unexpectedly on any change of the `config`.

kwinto:
I believe we need to reconsider our current approach to generation of the element ids. We are generating and reassigning them on every render cycle of the component while ideally you would do this only once.

kwinto:
> I believe we need to reconsider our current approach to generation of the element ids. We are generating and reassigning them on every render cycle of the component while actually you would do this only once.

We will tackle this in a separate task as this was not introduced in this pr.

kwinto (changes requested):
Fix focus jumping on unrelated changes of `config`

sushmi21:
> We have recently introduced the api to change webchat settings "on-the-fly" by calling webchat.updateSettings() method. You may want to remove `config` from the useEffect dependencies, because calling this method leads to the focus being shifted unexpectedly on any change of the `config`.

fixed!

kwinto (approved):

PR #174, opened 2021-04-13
Improve Carousel Accessibility

sushmi21:
This PR improves the accessibility of carousels.
Success criteria:
1. Carousel that has focus should be scrollable using arrow keys. This way, the user using a keyboard can view all cards, even if a card does not have an interactive element
2. Next Button of the carousel will not appear while tabbing out of the last item in the carousel. 
3. Each Carousel Item should have proper numbering indicating its position in the carousel

To test:
- Create Say node to include a gallery. Configure the gallery to include multiple cards with buttons. Also, include a card without a button in between other cards.
- Use the keyboard to move the focus to one of the items within the carousel. Now use right/left arrows to scroll through the list of cards.
- Move focus to the last button within the last card of the carousel. Now, on clicking Tab, the Next button should not appear. 
- On navigating the carousel using a screen reader, you should hear the list item position along with the title. (Example: list item 2 of 3: Some title)


src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -86,25 +87,55 @@ export const getMessengerGenericTemplate = ({
...
             const carouselRootId = `webchatCarouselTemplateRoot-${uuid.v4()}`;
             const carouselTitleId = `webchatCarouselTemplateTitle-${uuid.v4()}`;
             const carouselSubtitleId = `webchatCarouselTemplateSubtitle-${uuid.v4()}`;
+
+            const listItemCount = index !== undefined ? `${index + 1} of ${carouselListLength}` : undefined;
+            const carouselAriaLabel = title ? `${listItemCount}: ${title}` : listItemCount;
             const carouselAriaLabelledby = title ? carouselTitleId : undefined;
             const carouselAriaDescribedby = subtitle ? carouselSubtitleId : undefined;
-            const a11yProps = {role: "group", "aria-labelledby": carouselAriaLabelledby, "aria-describedby": carouselAriaDescribedby};
+            const carouselRooTA11yProps = {role: "group", "aria-labelledby": listItemCount ? undefined : carouselAriaLabelledby,
```
dshire, line 134: is this spelled correct? Or should the T be in lower case?

dshire (changes requested):
I feel like we've reached the point where we should split this component into smaller elements, but that is not the task here obviously.

Regarding the changes, it's working as described. However, I am not sure about the tab 
 selecting:
1. I can click on a carousel element's title and it is selected (grey background), but I can't select it like that using tab presses.
2. I can go left and right using the arrow keys, but if I went from carousel element 1 to element 3 using arrow keys, and then I want to use element 3's button, I can't just select it without starting from 1 again. 
I feel like I should be able to select 3's button quickly when I'm looking at 3. Maybe we can just make the carousel element titles selectable, that way you can just tab through them and through the buttons? Or maybe you know a better approach to this problem.

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
sushmi21: Thanks for pointing these out. I have pushed changes to fix those. 
1. You will not see the focused (grey background) style for elements that are not links
2. When a carousel element is brought into view using Arrow keys, you will now be able to focus the first interactive element within it, on clicking 'Tab'. 
3. Also, fixed the typo.

dshire (approved):

pedily:
On IE11, focusing a button element in another gallery item will scroll in a weird way, moving all gallery items out of reach.

recording

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -107,26 +124,58 @@ export const getMessengerGenericTemplate = ({
...
+                });
+            } else if(event.key === "ArrowLeft") {
+                this.setState({selectedItem: this.state.selectedItem - 1}, () => {
+                    this.focusCardInView();
+                })
+            }
+        }
+
+        focusCardInView = () => {
+            const nextCardToFocus = document.getElementById(`${this.carouselRootId}-${this.state.selectedItem}`);
```
pedily, line 156: either use refs or a scoped element search

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -107,26 +124,58 @@ export const getMessengerGenericTemplate = ({
...
         }
 
         renderElement = (element: IFBMGenericTemplateElement, index?: number) => {
             const { onAction, ...divProps } = this.props;
             const { image_url, image_alt_text, title, subtitle, buttons, default_action } = element;
 
+            const carouselListLength = this.props.payload.elements.length;
             const isCentered = this.props.config.settings.designTemplate === 2;
-            const carouselRootId = `webchatCarouselTemplateRoot-${uuid.v4()}`;
             const carouselTitleId = `webchatCarouselTemplateTitle-${uuid.v4()}`;
```
pedily, line 166: please change these so they won't generate a new id on every render
```
const [someId] = useState(() => uuid.v4())
```

pedily (changes requested):

sushmi21:
> On IE11, focusing a button element in another gallery item will scroll in a weird way, moving all gallery items out of reach.
> 
> recording

I have commented out the lines responsible for bringing the focused card into view. You will also not be able to scroll through the carousel using right/ left arrow keys. I have not removed the code as it is not clear why the issue occurs in IE11.

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
sushmi21: I don't think it will be an issue here, as the id is a random id generated using uuid. Moreover, this part does not have any effect now due to the commented out codes

PR #180, opened 2021-04-21
List template layout in IE11

Cesarcs7:
This PR fixes the layout of the webchat lists elements which was broken in IE11 since it does not support grid layout, I have changed it to flex

Please test the Lists templates and make sure they are looking and working as before in the webchat channel in IE11, also test in other browsers.

### DIFF
--- a/src/plugins/messenger/MessengerPreview/components/MessengerListTemplate/components/MessengerListTemplateElement.tsx
+++ b/src/plugins/messenger/MessengerPreview/components/MessengerListTemplate/components/MessengerListTemplateElement.tsx
@@ -73,7 +73,7 @@ export const getMessengerListTemplateElement = ({ React, styled }: MessagePlugin
                     tabIndex={default_action?.url ? 0 : -1}
                     onKeyDown={e => handleKeyDown(e, default_action)}
                 >
-                    <div style={{maxWidth:image_url ? '75%': ""}}>
+                    <div style={{maxWidth:image_url ? '74%': ""}}>
                         <MessengerTitle className="webchat-list-template-element-title" dangerouslySetInnerHTML={{ __html: title }} />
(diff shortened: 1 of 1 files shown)


src/plugins/messenger/MessengerPreview/components/MessengerListTemplate/components/MessengerListTemplateElement.tsx
```diff
@@ -44,43 +43,45 @@ export const getMessengerListTemplateElement = ({ React, styled }: MessagePlugin
...
             }
         }
 
         return (
             <div role="listitem">
                 <Root
-                    role={default_action?.url ? "link" : undefined} 
+                    role={default_action?.url ? "link" : undefined}
                     onClick={default_action && (e => onAction(e, default_action))}
+                    style={default_action?.url ? { cursor: "pointer" } : {}}
```
sushmi21, line 70: Merge regression: This inline style can be removed. The same style is applied using class 'link'.

Cesarcs7: fixed

sushmi21:
Previously, there was 8px gap between the columns of the grid (i.e. between the list item text and image). Now, we don't. If you have long list item titles, then it looks a little off. Please compare the two images.
Screenshot from 2021-04-22 17-49-26 Screenshot from 2021-04-22 17-55-21

Cesarcs7:
> Previously, there was 8px gap between the columns of the grid (i.e. between the list item text and image). Now, we don't. If you have long list item titles, then it looks a little off. Please compare the two images.
> Screenshot from 2021-04-22 17-49-26 Screenshot from 2021-04-22 17-55-21

I fixed this the closer i could to make it look like the behaviour with the grid before.

sushmi21 (approved):
Looks good to me

pedily (commented):
works fine :)

PR #181, opened 2021-04-29
Add opt-in markdown styling for regular text messages

pedily:
This PR adds the `enableMarkdownStyling` flag which will apply Markdown-specific CSS styling to the content of regular text messages.

To test this, set the flag and let a flow output the content of the example HTML output of [markdown-it](http://markdown-it.github.io/) (which is used in the background).


kwinto:
It effects user message bubble (white text on white background):
Screenshot 2021-04-29 at 23 39 32

kwinto:
The stylings in general with flag enabled make a lot more sense in comparison with default ones.

Thinking, should we rename the flag to something more general (maybe enableHtmlStyling?) and apply this styles by default with next major release?

kwinto:
Anchor hover colour lacks contrast
Screenshot 2021-04-29 at 23 52 22

kwinto:
Should we apply display:block on table content tags? May have an unexpected effect for the user but seems better to me than such squeezed text
Screenshot 2021-04-30 at 00 04 20

src/webchat-ui/components/presentational/MarkdownMessageBubble.tsx
```diff
@@ -0,0 +1,132 @@
...
+      margin-top: 0;
+    }
+  
+    &>*:last-child {
+      margin-bottom: 0;
+    }
+  
+    pre,
+    code {
+      font-family: Menlo, Monaco, "Courier New", monospace;
```
kwinto, line 38: Missing some common windows and linux fonts:
```
font-family: ui-monospace, Menlo, Monaco, "Cascadia Mono", "Segoe UI Mono", "Roboto Mono", "Oxygen Mono", "Ubuntu Monospace", "Source Code Pro", "Fira Mono", "Droid Sans Mono", "Courier New", monospace;

dshire:
> Should we apply display:block on table content tags? May have an unexpected effect for the user but seems better to me than such squeezed text
> Screenshot 2021-04-30 at 00 04 20

Don't think we should do that, users should rather style their webchat in a way that allows for a large table if that is what they are going for. Wouldn't display block basically break the table layout and display things from the same rows in different ones?

dshire:
Maybe change it, so that links are only underlined on hover. I think that is more common and readable.

dshire:
Screenshot from 2021-04-30 10-23-28
Screenshot from 2021-04-30 10-23-24
I guess they're not that important, but the definition lists styling is a bit off, e.g. missing the "bold" style for the term

dshire:
Screenshot from 2021-04-30 10-24-46
Screenshot from 2021-04-30 10-24-38
Code could look cleaner. Reduce the font size for code blocks, maybe even just use the classic grey background with black color letter?
Also, inline code is styled same as a code block, maybe change it a bit like in the example? At least the dark background needs some padding. 
Syntax highlighting is not working (I guess that would need a plugin?)

kwinto:
> Maybe change it, so that links are only underlined on hover. I think that is more common and readable.

I would left links underlined. Otherwise you would need a new theme colour specifically for links to stand out

dshire:
> > Maybe change it, so that links are only underlined on hover. I think that is more common and readable.
> 
> I would left links underlined. Otherwise you would need a new theme colour specifically for links to stand out

Yes that's what I meant with my suggestion, give them the different color they get on hover always, and add the underline when hovered (that is how the markdown-it demo displays it). Currently it's the other way around. (Or just leave the underline out completely and just change the shade of the link color on hover. That is for example how links are displayed in the visual studio wiki.)

pedily:
Regarding the "link highlighting":
We cannot set a "fixed color" here since the text color is already the "best contrast color" to whatever theme is active right now, so we cannot go further than that.
I would also not go for "different font styling" since that implicitly overrides bold, italic and underlined.
We can play around with different text decorations or text shadows.

dshire (approved):

kwinto (approved):

PR #186, opened 2021-05-10
Fix carousel accessibility

sushmi21:
This PR fixes the accessibility issues in the carousel component when used in (relatively new) browsers. 

**Success Criteria:**
1. When an element within the carousel is focused, you should be able to scroll through the carousel cards using right and left arrow.
2. When 'Tab' key is pressed to navigate through the clickable elements in the carousel, the card that has the focused element should be brought into view,
3. When tabbing out of the last interactive element present in the last carousel card,  the focus should not move to the 'Next Item'  button of the carousel.
4. When using a screen reader, the card number that has the focus should be announced. 
5. On keyboard navigation, the carousel should not scroll out of view causing weird behavior (See attached video for possible weird behavior).
6. Steps 1 to 4 will **not** be true in IE 11 and Edge 15 - 18. (These issue were not newly introduced in this PR or any of the accessibility fixes PRs). Step 4 must be true in all browsers.

https://user-images.githubusercontent.com/29442479/117649046-53d77180-b18f-11eb-958e-ede5ff231102.mp4

How to test:
1. Add a Say node in default channel and configure it to include a 'gallery'.
2. Add multiple cards. Include at least one button in some of the gallery cards.
3. Interact with the bot and use keyboard navigation to move focus to one of the carousel items.
4. Use Right/ Left arrow keys to horizontally scroll the carousel.
5. When a card is brought into view, press 'Tab' and see if the first interactive element in the current card is focused.
6. Use 'Tab' only navigation and see if all interactive item receiving focus is in view. Doing this should not move the carousel completely out of view.
7. Test the above behavior also in IE11 and Edge 18. Step 4 and 5 will **not** work in these browsers
8. Also, use a screen reader to test the above mentioned success criteria.


cbuchwald (approved):
Tested in IE11, Edge 15, Edge 90, Firefox 88 and Chrome 90, worked as expected :)

Cesarcs7 (approved):
lgtm

PR #192, opened 2021-05-27
Feature/11814 open a url in the existing tab

cbuchwald:
Description:

This PR enables the Webchat Widget to open URLs from buttons, galleries and lists in the same tab using the "target" property that can be applied in Cognigy AI version 4.4.0+

Success criteria:

- The URL opens in the same tab

How to test (steps):

1. Create a say node with a button with target self
2. Click the button, it should open in the same tab
3. Create a say node with a button with no target defined
4. Click the button, it should open in a new tab
5. Test this behaviour in galleries with buttons and lists with a button

Additional considerations (describe):

- [ ] This PR impacts NLU
- [ ] This PR might have performance implications
- [ ] This PR changes an existing data model (and might affect existing legacy data)
  - Examples: adding, renaming, removing a field or changing the format or constraints of a field

Security implications (describe):

- [ ] Possible injection vector
- [ ] Authentication/Access controls touched
- [ ] Sensitive Data could be exposed
- [ ] XSS
- [ ] Logging/Monitoring touched
- [ ] Exchanges data with external systems
- [x] No security implications

### DIFF
--- a/src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
+++ b/src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
@@ -37,7 +37,8 @@ export const getMessengerButton = ({ React, styled }: MessagePluginFactoryProps)
         const buttonType = button.type;
         const isWebURL = buttonType === "web_url";
         const buttonTitle = button.title ? button.title + ". " : "";
-        const ariaLabel = isWebURL ? buttonTitle + "Opens in new tab" : undefined;
+        const isWebURLButtonTargetBlank = button.target !== "_self";
+        const ariaLabel = isWebURL && isWebURLButtonTargetBlank ? buttonTitle + "Opens in new tab" : undefined;
 
         return (
             <Button 
--- a/src/plugins/messenger/index.tsx
+++ b/src/plugins/messenger/index.tsx
@@ -80,7 +80,9 @@ const messengerPlugin: MessagePluginFactory = ({ React, styled }) => {
                     // @ts-ignore
                     if (action.type === 'web_url' && action.url) {
                         // @ts-ignore
-                        window.open(action.url, '_blank');
+                        const target = action.target === "_self" ? "_self" : "_blank";
+                        // @ts-ignore
+                        window.open(action.url, target);
                     }
                 }}
                 config={config}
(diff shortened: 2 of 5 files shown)


src/plugins/messenger/index.tsx
```diff
@@ -80,7 +80,9 @@ const messengerPlugin: MessagePluginFactory = ({ React, styled }) => {
                     // @ts-ignore
                     if (action.type === 'web_url' && action.url) {
                         // @ts-ignore
-                        window.open(action.url, '_blank');
+                        const target = action.target === "_self" ? "_self" : "_blank";
+                        // @ts-ignore
+                        window.open(action.url, target);
```
sushmi21, line 85: We could simply pass `action.target`, right? 
```suggestion
                        window.open(action.url, action.target);
```

cbuchwald: Target needs to be "_self" or "_blank" otherwise it will run into errors. I tried it like this first, but then changed it to get sure it's always set to one of these values. This should be the best approach as previous versions of Cognigy AI didn't send a target property and would therefore become incompatible.

cbuchwald: Ah, now I get it...yes sure, I can modify it so we don't need another variable :)

sushmi21:
looks good to me. Only thing we need to take care now is the '_Opens in new tab_' aria-label for links. We need to now remove this aria-label for links opening in the same tab. I can take care of this in a separate PR.

sushmi21 (approved):
TODO: Fix aria-labels for links opening in same tab

pedily:
@sushmi21 I think we should directly handle the aria-labels here, right? Otherwise we have a regression. Can you two fix that together?

cbuchwald:
@pedily sure we'll take care of it

sushmi21:
> @sushmi21 I think we should directly handle the aria-labels here, right? Otherwise we have a regression. Can you two fix that together?

Done! Just needed to fix it in one place

src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
```diff
@@ -37,7 +37,8 @@ export const getMessengerButton = ({ React, styled }: MessagePluginFactoryProps)
         const buttonType = button.type;
         const isWebURL = buttonType === "web_url";
         const buttonTitle = button.title ? button.title + ". " : "";
-        const ariaLabel = isWebURL ? buttonTitle + "Opens in new tab" : undefined;
+        const isWebURLButtonTargetBlank = button.target !== "_self";
+        const ariaLabel = isWebURL && isWebURLButtonTargetBlank ? buttonTitle + "Opens in new tab" : undefined;
```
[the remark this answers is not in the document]
sushmi21: Hmm..if you look at line 39, I concatenate the button title with a dot and space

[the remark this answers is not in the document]
kwinto: Ok, good!

pedily (approved):

PR #198, opened 2021-07-06
bug/15533-Newlines-in-Content

cbuchwald:
Fixed whitespace for multiline text with wrap

This solves: Issue that say nodes with multiline text were displayed in one line

How to test: Create a flow with a say node with multiple lines of code and view it using the modified webchat. Multiline Text should be displayed in several lines


Cesarcs7 (approved):

sushmi21:
It seems to work fine for me even without this change. Also, the file name says 'MarkdownMessageBubble'. So, I am not sure if this is the right file.

kwinto:
lgtm @sushmi21 

this template is used when Webchat is initialised with `enableGenericHTMLStyling: true`. It makes sense to have it in sync with regular template regarding the white-space behaviour.

sushmi21:
> lgtm @sushmi21
> 
> this template is used when Webchat is initialised with `enableGenericHTMLStyling: true`. It makes sense to have it in sync with regular template regarding the white-space behaviour.

Okay, then will approve. thanks :)

sushmi21 (approved):
It worked for me even if I revert this change. But, will approve this anyway since Dmitrii reconfirmed that this won't break anything else. :)

PR #199, opened 2021-07-21
Focus input field after clicking postback or quick reply

sushmi21:
This PR adds a new feature flag which when true will move focus to the input field after a quick reply or a post back button is clicked.

To test: 
- Use a flow that has quick reply buttons and postback buttons.
- In index.html, set `focusInputAfterPostback: true`
- In the chat, if you now click postback or quick reply button, focus will move to input text.
- Change the flag to false and see if the focus doesn't change


cbuchwald (approved):

dshire (changes requested):
Works fine for must buttons etc. but not for buttons on gallery cards. as these are also postback values, it should work the same

sushmi21:
> Works fine for must buttons etc. but not for buttons on gallery cards. as these are also postback values, it should work the same

Fixed. Thanks!

dshire (approved):
lgtm

PR #205, opened 2021-08-06
fix rating dialog position for old browsers

dshire:
dialog position in IE11 was messed up, fixed by adding top: 0 and left: 0 for the absolute positioning


sushmi21 (approved):
Works as expected in IE11 as well

PR #207, opened 2021-08-09
Separate Build + Compatibility expansion for "rating plugin"

pedily:
This PR
- adds a separate build step for building the "rating plugin" (we now use it in our UI service as well)
- adds compatibility to the "message with rating"-format we use in our backend (or: in our message histories in our UI)
- replaces a styling instruction with an identical other instruction so it's rendered correctly in our UI


src/plugins/rating/RatingHistoryEntry.tsx
```diff
@@ -38,16 +35,16 @@ const RatingHistoryEntry = styled.div(({ theme }) => ({
...
 
 	const ratingText = config.settings.ratingMessageHistoryRatingText;
 	const commentText = config.settings.ratingMessageHistoryCommentText;
 
 	if (!rating) return null;
 
 	const getRatingIcon = (rating: number) => {
-		if (rating === 10) return <StyledThumbIcon />;
-		if (rating === 0) return <StyledThumbDownIcon />;
+		if (rating === 10) return <StyledThumbIcon style={{ marginLeft: 6 }} />;
```
dshire, line 46: would be nicer to move inline styling to styled component (adding it once in line 11 should be enough)

dshire (approved):
code looks fine, rating feature is still working as before. you could move the inline style to the styled component.

sushmi21 (approved):

src/plugins/rating/RatingHistoryEntry.tsx
sushmi21: @pedily Will you be addressing this? Else, I will merge the PR.

PR #209, opened 2021-08-10
Bug/16812 a11y regression chat rating dialog

sushmi21:
This PR fixes the accessibility issues in rating dialog.

**Success Criteria:**

- The rating thumbs up/down button should have aria label
- All the buttons in the dialog should have aria-labels
- The text field in the dialog should have an associated label
- All focusable elemnents should receive 'visible' focus
- You should be able to navigate the dialog using keyboard
- You should be able to close the dialog by pressing `Escape` key
- When the dialog is open, the focus should not leave the dialog on pressing `Tab `or `Shift + Tab`
- When the dialog is closed, the focus should go back to the thumbs up/down button
- Focus should not leave the webchat widget when 'enableFocusTrap' feature is enabled. But, you should still be able to move focus from one element to another in meaningful order.
- When both `enableRating` and `enableFocusTrap` flags are set to true, the focus order on `Tab` and `Shift + Tab` navigation should be meaningful. 
- When using a screen reader, the rating dialog and the elements within should be properly announced

**How to test:**

Please read the success criteria and test if everything is achieved

- Set `enableRating` flag to true to enable the rating button in the chat header
- Use your keyboard and test if you could open, navigate and close the dialog. 
- Navigating the dialog using Tab or Shift + Tab, and check if the focus does not leave the dialog
- Check the dialog accessibility using a screen reader. (For instance: ChromeVox)
- Now also set `enableFocusTrap` to true. 
- Navigate the chat window using Tab and Shift Tab, the focus should not leave the chat window and the focus order should be meaningful


dshire (changes requested):
looks good, I only found two things when using the screen reader:
1. The send button in the rating dialog just reads "button"
2. The rating entry in the history does not read the "the thumbs up/down"-icons in any way

sushmi21:
> looks good, I only found two things when using the screen reader:
> 
> 1. The send button in the rating dialog just reads "button"
> 2. The rating entry in the history does not read the "the thumbs up/down"-icons in any way

Good point! I did not make the rating entry in the message history accessible. Totally forgot about that. Will fix them soon

sushmi21:
> looks good, I only found two things when using the screen reader:
> 
> 1. The send button in the rating dialog just reads "button"
> 2. The rating entry in the history does not read the "the thumbs up/down"-icons in any way

Fixed! thanks :)

dshire (approved):
lgtm

Cesarcs7 (commented):
Works well, but I found that the layout displays a horizontal scrollbar. Was this introduced in this PR?
Screenshot from 2021-08-13 12-10-08

sushmi21:
> Works well, but I found that the layout displays a horizontal scrollbar. Was this introduced in this PR?
> Screenshot from 2021-08-13 12-10-08

In which browser was this? I don't see this in Chrome

Cesarcs7 (approved):
We realized that the breaking layout was because of a quick reply component. So I will approve.

PR #210, opened 2021-08-12
Feature/16366 allow javascript content

pedily:
This PR introduces sanitization options for HTML content in messages as well as the URLs of URL buttons and "default actions".

They are enabled by default and can be disabled by setting a flag


Cesarcs7 (approved):

src/plugins/messenger/MessengerPreview/components/MessengerGenericTemplate/MessengerGenericTemplate.tsx
```diff
@@ -240,6 +244,7 @@ export const getMessengerGenericTemplate = ({
 											button={button}
 											onClick={e => onAction(e, button)}
 											className="webchat-carousel-template-button"
+                                            config={this.props.config}
```
sushmi21, line 247: This particular line is indented using spaces, but the wrapper element is indented using Tabs. We need to find a solution for this. We could see the same issue in other PRs too.

sushmi21 (approved):

PR #211, opened 2021-08-19
feature/16692 add logic to use QuickReply title as postback

cbuchwald:
Description:

This PR adds a functionality that clicked QuickReply buttons return the title instead of the payload in case the useQuickReplyTitleAsPostback option is set to true 

Success criteria:

- QuickReplies return the title as postback if the flag is set

How to test (steps):

1. Embed the webchat with the setting useQuickReplyTitleAsPostback enabled
2. Create a flow with a say node with a quick reply with a title other than the postback
3. Create another say node with a Messenger channel QuickReply with a title different than the payload
4. Check the response in another saynode sending the response {{ci.text}} it should return the title instead of the payload

Additional considerations (describe):

- [ ] This PR impacts NLU
- [ ] This PR might have performance implications
- [ ] This PR changes an existing data model (and might affect existing legacy data)
  - Examples: adding, renaming, removing a field or changing the format or constraints of a field

Security implications (describe):

- [ ] Possible injection vector
- [ ] Authentication/Access controls touched
- [ ] Sensitive Data could be exposed
- [ ] XSS
- [ ] Logging/Monitoring touched
- [ ] Exchanges data with external systems
- [x] No security implications


src/webchat/store/messages/message-middleware.ts
```diff
@@ -93,6 +93,17 @@ export const createMessageMiddleware = (client: SocketClient): Middleware<{}, St
             const isMessageEmpty = !(message.text || message.data?._cognigy?._webchat);
             const isUnseen = !isWebchatActive && !isMessageEmpty;
 
+            if(store.getState().config.settings.useQuickReplyTitleAsPostback) {
```
pedily, line 96: this should rather be handled in the messenger plugin rather than at a central place.

pedily (changes requested):

src/webchat/store/messages/message-middleware.ts
cbuchwald: Thx, fixed it

src/plugins/messenger/MessengerPreview/components/MessengerTextWithQuickReplies/MessengerTextWithQuickReplies.tsx
```diff
@@ -102,6 +102,7 @@ export const getMessengerTextWithQuickReplies = ({
                                     const { title, image_url, image_alt_text } = quickReply as IFBMTextQuickReply;
                                     label = title;
                                     if (image_url) image = <QuickReplyImage src={image_url} alt={image_alt_text || ""}/>;
+                                    if (config.settings.useQuickReplyTitleAsPostback) (quickReply as IFBMTextQuickReply).payload = title;
```
pedily, line 105: Instead of modifying the object, please just create a variable that is either set to one or the other value and use it there

pedily (changes requested):

src/plugins/messenger/MessengerPreview/components/MessengerTextWithQuickReplies/MessengerTextWithQuickReplies.tsx
cbuchwald: thx, fixed it

pedily (approved):

sushmi21 (approved):

PR #215, opened 2021-08-31
changed rating feature to use -1/1 instead of 0/10

dshire:
This PR changes the rating feature to use values of -1/1 instead of 0/10


pedily (approved):
works as expected

sushmi21 (approved):

PR #218, opened 2021-09-28
added "disableInputAutocomplete" flag

pedily:
This PR lets you disable the "autosuggest" feature from the regular text input field following bug reports that Samsung phone keyboards tend to display the autosuggestion options above the software keyboard leading to errors.


pedily:
was able to reproduce the issue with "false" and kill it with "true"!

PR #219, opened 2021-09-30
Set up automatic browser compatibility testing via Github Actions > BrowserStack Automate > Selenium Webdriver

dshire:
This PR adds the automatic browser compatibility testing I've prepared in a fork. For reference, please refer to the internal communication regarding this and/or reach out to me.

Note: This only works with the browserstack secrets mentioned in the communication


dshire:
Let's please Squash merge this once it's ready, there are way too many commits that noone will ever need to read :)

PR #223, opened 2021-11-04
Fix crash when buttons are "null"

pedily:
This PR fixes an issue from a customer where our "messenger templates" wouldn't handle "null" as a value for the "buttons array" well and the webchat would crash.

The solution is to assume that the buttons value is an empty array in case it's falsey.
I also checked quick replies and added tests.

See the test files for clarification.


dshire (approved):
lgtm

EgleHelms (approved):

PR #231, opened 2021-12-16
Fix race condition in "persistent storage" when using useSessionStorage

pedily:
This PR fixes an issue where users of the `useSessionStorage` flag could run into a race condition which caused the persistent history to be lost.

When the webchat initializes, it loads the config from the endpoint URL, then merges the endpoint config with the overrides and applies it to the webchat configuration. Until then, the "default" settings applied.

If the "restore persistent history" mechanism triggered before the webchat config was resolved from the endpoint, it would always restore from `localStorage`, regardless of the `useSessionStorage` flag applied in the embedding settings, because those embedding settings would only be applied after the endpoint config was resolved.

The result would be that we would restore from an empty history (because we restored from `localStorage` instead of `sessionStorage`, and there was no content) and then we would start writing our content to the correct location, overriding any previous content.

To mitigate this, I added a new action that applies the "embedding override settings" directly on mount in addition to "when the config was resolved from the endpoint".

This will always happen before we try to restore the persistent history.
As long as the `useSessionStorage` option is set via embedding parameters, it will be respected properly.

To test this issue, I had to upgrade cypress to a more recent version since they extended their HTTP request interceptor capabilities (...which I used for adding an artificial delay to the response).

I also had to touch 2-3 other test files where we used assertions on non-existent DOM elements which would fail from cypress v6.0.0 onwards.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,6 +1,5 @@
 {
     "baseUrl": "http://localhost:8787/",
     "video": false,
-    "experimentalNetworkStubbing": true,
     "chromeWebSecurity": false
 }
--- a/cypress/integration/engagement.spec.ts
+++ b/cypress/integration/engagement.spec.ts
@@ -42,7 +42,7 @@ describe('Engagement Message', () => {
             .wait(500)
             .window()
             .contains('engagement message text', { timeout: 500 })
-            .should('not.be.visible')
+            .should('not.exist');
     });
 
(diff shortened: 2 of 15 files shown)


kwinto:
New test fails when I comment out relevant changes. 

Nice PR :)

kwinto (approved):

cbuchwald (approved):

PR #239, opened 2022-01-09
feature/20924 upgrade react and emotion

cbuchwald:
How to test:

Run all tests and check if all plugins still work.

@emotion/cache is still v10 because v11 has breaking changes to the plugin system


EgleHelms (approved):

pedily:
We also need to validate that the plugin outputs still work in the UI / interaction panel.

package.json
```diff
@@ -1,112 +1,109 @@
...
+		"@types/dompurify": "^2.0.0",
+		"@types/react": "^16.9.11",
+		"@types/react-dom": "^16.9.4",
+		"@types/react-redux": "^6.0.12",
+		"@types/tinycolor2": "^1.4.3",
+		"@types/uuid": "^3.4.6",
+		"@types/whatwg-fetch": "0.0.33",
+		"babel-loader": "^8.2.3",
+		"babel-polyfill": "^6.26.0",
+		"buffer": "^6.0.3",
```
pedily, line 84: why was this package added?

pedily: I was able to build and test successfully without it

pedily:
We also definitely need to announce the `react` and `styled` upgrade as a "potentially breaking change" in our release notes since they are forwarded to plugin builders

pedily (commented):
I ran the tests and manually executed the webchat on modern firefox and IE11 and found no issues.

Please check whether you can get rid of the `buffer` package. When we merge this, we should do a webchat release and explicitly line out the upgrades to plugin developers

webpack.config.js
```diff
@@ -11,7 +12,11 @@ module.exports = {
         filename: 'webchat.js'
     },
     resolve: {
-        extensions: ['.ts', '.tsx', '.js', '.json']
+        extensions: ['.ts', '.tsx', '.js', '.json'],
+        alias: {
```
pedily, line 16: Is this because of the `@emotion/cache` package? Otherwise, I think we should prefer changing our imports rather than setting up an alias

cbuchwald: When testing it in my MBO I needed to have it, now it seems to work without, I guess it was because of @emotion/cache. I removed it.

package.json
cbuchwald: True, seems to be a leftover from testing, works fine without, removed it

cbuchwald:
> I ran the tests and manually executed the webchat on modern firefox and IE11 and found no issues.
> 
> Please check whether you can get rid of the `buffer` package. When we merge this, we should do a webchat release and explicitly line out the upgrades to plugin developers

removed

PR #240, opened 2022-01-10
Add default support for "adaptivecards"

pedily:
This PR adds support for AdaptiveCards by default.

The code is a slightly updated version of our [AdaptiveCards Plugin](https://github.com/Cognigy/WebchatPlugins/tree/master/plugins/adaptivecards), where the outdated [`react-adaptivecards`](https://www.npmjs.com/package/react-adaptivecards) package was replaced with a custom implementation based on Microsoft's [`adaptivecards-react`](https://www.npmjs.com/package/adaptivecards-react). The latter one seems to be not yet ready for a stable release as it depends on a beta-version of [`adaptivecards` v3](https://www.npmjs.com/package/adaptivecards).

I added the card from the example page of the Webchat Plugin and two tests to validate that both rendering and form submission are working correctly.
To test the submission, I added a cypress utility plugin that lets you "wait for anything" like cypress does for `.get()` and `.contains()`, used here to "wait for an object in the message history" because the submit message wouldn't be visible.

The addition of adaptivecards affects the file size  both the uncompressed `webchat.js` and `webchat.legacy.js` by ~230KB.


cypress/integration/messages/adaptivecard.spec.ts
```diff
@@ -0,0 +1,36 @@
+/// <reference path="../../support/index.d.ts" />
+
+describe("Message with Audio", () => {
```
cbuchwald, line 3: Message with Adaptive Card

cbuchwald:
Where is the documentation mentioned in the user story?

cbuchwald (approved):

sushmi21 (approved):
Works as expected

PR #241, opened 2022-01-13
feature/20929 refactor and move date picker and moment

cbuchwald:
This PR deletes the local copied react-flatpickr and reuses it via NPM package. It replaces Moment with Intl in the code and adds some tests to verify the date format is still the same.

Moment is now included as a dev dependency for comparing the outputs, reducing the uncompressed file size of date-picker.webchat-plugin.legacy.js from 477,8kb to 159,5kb and a compressed version with 30,3kb.

How to test:
1. Create a flow with a date picker
2. Test each configuration option
3. run "npm run test"
4. Everything should still work the way it did before


pedily:
Please check all the available options we have in the UI for the date picker in a modern browser and in IE11.
I found that `zh_tw` is not showing a date picker in the new library, it might need an alias.

pedily:
We have to double-check that IE11 supports all of the locale formats configurable in the UI.
Date.js supports adding additional locales as an opt-in, and in case we need them for IE11, we can add them "only for the legacy build"

pedily (changes requested):
please check the compatibility with our predefined locale options

PR #244, opened 2022-01-25
Added webchat.clearMessages() function

mbrau12s:
One of our customers requested this feature, would be nice if you could merge it! :)


pedily:
Hi @mbrau12s,
Thank you for actively contributing to the WebchatWidget!

I'm happy to add this functionality, but we'll need to do the following additional things first:
- the "API Documentation" should be updated as well featuring a small explanation and a code example (`docs/webchat-api.md`)
- to avoid breaking this, we should add automated tests which check that
  - the history is cleared when the webchat is open
  - that the "unread message badge" and "unread message preview" are cleared when the webchat is not open and has unread messages
  - the history stays cleared after the page reloads using the same userId and sessionId

If you want, I can help you with the automated tests. Just let me know :)

dheckl:
Is it possible to get the requirements from @pedily done and the PR merged?

PR #247, opened 2022-01-28
add prettier config file and "prettier:check" script

pedily:
This PR adds a prettier config and a check command as a starting point for an automated prettify-checking approach.
As a follow-up, when there's a point in time without work-in-progress on the webchat, we will create a PR which prettifies all code and auto-checks prettification status as a PR criteria


cbuchwald:
According to the prettier documentation it should automatically read .prettierrc.json, for me it didn't work out of the box, therefore we should hardcode it with `"prettier:check": "prettier --check --config .prettierrc.json src/"` instead.

pedily:
@cbuchwald thanks, applied that!

EgleHelms (approved):
The check works.

cbuchwald (approved):

PR #251, opened 2022-02-15
Feature/21499 Improve aria-label for button groups

sushmi21:
This PR adds total number of buttons / links in a button or quick reply, to the group's aria-label.

Success criteria:
- When there is more than one button/link in a quick reply or text with buttons group, you should hear the screen reader announcing total number of buttons/ links.
- When there is only one button, there should be no change in the behavior.
- No layout regressions

How to test:
- Create a flow that has say nodes with output types Text with Buttons and Text with quick replies. Include more than one buttons or links. Alternatively, you can use this endpoint [https://endpoint-dev-v4.cognigy.ai/39d8ff845d0469d85be1c6d9b01cb435803cfede1b7dfb782e74f9343ec27fc7](https://endpoint-dev-v4.cognigy.ai/39d8ff845d0469d85be1c6d9b01cb435803cfede1b7dfb782e74f9343ec27fc7)
- Chat with the bot and use the screen reader while navigating through the messages and test the above success criteria

### DIFF
--- a/cypress/integration/messages/buttons.spec.ts
+++ b/cypress/integration/messages/buttons.spec.ts
@@ -36,11 +36,21 @@ describe("Message with Buttons", () => {
                 .get(".webchat-buttons-template-root")
                 .get(".webchat-buttons-template-button")
         })
-	})
-	
-	it("group off buttons should have role 'group'", () => {
+    })
+
+    it("group off buttons should have role 'group'", () => {
         cy.withMessageFixture('buttons', () => {
-			cy.get("[role=group] .webchat-buttons-template-button").should("have.length", 4);
-        })				
+            cy.get("[role=group] .webchat-buttons-template-button").should("have.length", 4);
+        })
+    })
+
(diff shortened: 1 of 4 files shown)


pedily:
For button or quick reply messages with only one button or quick reply, the screenreader does not announce the "context". For other messages, it says that we're in a message with text XYZ

pedily:
If this is something we should avoid breaking then I'd be happy to see some tests checking for the texts to be available and be shaped correctly.

pedily (commented):
Please check whether the skipping of announcements for messages with only one button are intended or not.

I'd be happy to see some automated tests to support this :)

sushmi21:
> Please check whether the skipping of announcements for messages with only one button are intended or not.
> 
> I'd be happy to see some automated tests to support this :)

It is the expected behavior. We are adding the message as aria-label for the group of buttons. When there is only one button, there is no grouping and hence the label is also not added. 

I have added some e2e tests for this. Please re-review it :)

pedily (approved):

PR #252, opened 2022-02-22
bug/21904 get content from pointer

cbuchwald:
This PR fixes a bug from a customer causing the Webchat Widget to crash if quick replies are hidden behind a pointer which might happen if an intent has a default reply with quick replies.

How to test:
1. Create a flow with a say node, add the fixture data object to the data payload and try it with a previous version of the webchat widget, it should break.
2. Try it with this version, it should work.

Added tests as well

### DIFF
--- a/src/plugins/messenger/index.tsx
+++ b/src/plugins/messenger/index.tsx
@@ -8,6 +8,7 @@ import { IMessage } from "../../common/interfaces/message";
 import { IWebchatConfig } from "../../common/interfaces/webchat-config";
 import { sanitizeUrl } from "@braintree/sanitize-url";
 import { IFBMURLButton } from "./MessengerPreview/interfaces/Button.interface";
+import { transformPointers } from "./helpers";
 
 const getMessengerPayload = (message: IMessage, config: IWebchatConfig) => {
     const cognigyData = message.data?._cognigy;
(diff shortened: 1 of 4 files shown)


docs/embedding.md
```diff
@@ -122,6 +122,7 @@ See it in action:
...
 | unreadMessageTitleTextPlural       | string                              | "New Messages"                     | x                             |                                       |                                 | The website title that is displayed when the user retrieved more than one new message                                                                                                                           |
+| useDefaultReplyCompatibilityMode       | boolean                              | false                 |                              |                                       |                                 | A hotfix to avoid breaking the Webchat Widget when using default replies in intents. Enable it to avoid breaking, but be aware that it might decrease the performance and it will be removed in upcoming versions when the main issue is solved in Cognigy.AI                                                                                                                            |
```
pedily, line 125: please use the `enableXYZ` naming convention here

src/plugins/messenger/index.tsx
```diff
@@ -22,10 +25,10 @@ const getMessengerPayload = (message: IMessage, config: IWebchatConfig) => {
             return _facebook;
         }
 
-        return _webchat;
+        return transformPointers(_webchat,useDefaultReplyCompatibilityMode);
```
pedily, line 28: Now the code will always run through the "hotfix" code which can handle "being disabled" in addition.

I'd prefer to "pick the right object" first and then conditionally apply the `transformPointers` function on top of it. You can make use of IIFEs to get the data object first before deciding whether to return or transform it.


```javascript
() => {
  const theCorrectDataObject = (() => {
    // all the "pick the right object" logic goes here
  })();

  if (compatibilityMode)
    return transform(theCorrectDataObject);

  return theCorrectDataObject;
}
```

pedily (changes requested):

PR #256, opened 2022-03-04
Add e2e tests for the "rating" functionality

pedily:
This PR adds automatic tests for the "rating" functionality.
Since we're handling raw socket events outside of `RECEIVE_MESSAGE` for that, I modified the `messageHandler.ts` code to expose a `createOutputHandler` function and expose an instance of that via `webchat._handleOutput`.

As this is an "internal" API, we won't document it.

The `cy.receiveMessage` method now uses `webchat._handleOutput` instead of `webchat.store.dispatch(/* RECEIVE_MESSAGE EVENT PAYLOAD */);

I wasn't able to find an issue with the "rating" functionality (which was the original goal to fix in this PR).

### DIFF
--- a/cypress/integration/rating.spec.ts
+++ b/cypress/integration/rating.spec.ts
@@ -1,0 +1,214 @@
+describe("Rating", () => {
+    beforeEach(() => {
+        cy.visitWebchat();
+    })
+
+    it("header button shouldn't show up by default", () => {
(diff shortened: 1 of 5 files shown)


EgleHelms:
Works well for me. Both the functionality of Rating and the spec file is working as expected.

cbuchwald (approved):

PR #260, opened 2022-03-28
Feature/21501 date picker a11y fixes

sushmi21:
This PR fixes accessibility issues of a date picker (without time picker).

Success criteria:
- Navigating the date picker using 'Tab' and 'shift + tab' should trap the focus within the date-picker
- Using Left/ Right arrows should move focus to previous/ next day.
- Using Up/ Down arrows should move focus to the same day of previous/ next week.
- Using Control + Left/ Right arrows should change the grid of dates to previous/ next month.
- Using Control + Up/ Down arrows should change the grid of dates to previous/ next year.
- Should be able to close the date picker using Escape key press
- Check if Screen reader reads the label and description of date picker when you open the picker (Chrome-Vox might not read the aria-describedby value sometimes)
- Check if you are able to submit your selection
- Selecting single, multiple and range using keyboard should be possible

How to test:
- Use keyboard and Screen reader to test the above mentioned success criteria


sushmi21:
> In IE11, I was not able to select a date prior to the "5th of april" using the keyboard even though dates from the "1st of april" should have been available.
- I will check this now.

> Using the "up" and "down" arrow keys caused a diagonal focus change instead of a vertical one in IE11.
- Yes, I noticed that too. The library is handling the arrow key navigation. 

> I can only start moving using arrow keys if I move horizontally first (tested in Firefox and IE11). Is that intended?
- Yes, This is one of the limitations. We can investigate this in a follow-up

PR #267, opened 2022-04-29
Fix "sideloaded plugins" not taking preference over "internal plugins"

pedily:
This PR fixes an issue where "external plugins" would not win over "internal plugins".
The problem here was that with `v2.40.0`, we introduced an "internal" `adaptivecards` plugin.

For customers that loaded the existing `adaptivecards` plugin, that external plugin would not be used.
Anything apart from feature parity between the plugins would be recognized as a regression.
As we upgraded the "internal plugin" due to outdated dependencies in the "old" one, there were imparities reported as regressions.

While we put work into tackling those imparities, we also want to ensure that you can "choose" at any time, thus fixing the "loading priority" issue.


cbuchwald (approved):

EgleHelms:
@pedily please resolve an existing conflicts :)

EgleHelms (changes requested):
Please resolve a conflict.

pedily:
@EgleHelms done :)

EgleHelms (approved):

PR #270, opened 2022-05-11
Add button count in group for screen reader

sushmi21:
This PR adds to position of a button in a group of buttons to help screen reader users.

Success Criteria:
- In addition to the total number of buttons, the screen reader should also announce the position of the button.
Example: If a quick reply has three buttons, focusing on the second button should read '_Item 2 of 3 : <BUTTON_LABEL> button_'
- Links that open in a new tab should still have an aria-label that includes 'Opens in a new tab' 

How to test: 
- You can use this endpoint to test this PR: https://endpoint-dev-v4.cognigy.ai/39d8ff845d0469d85be1c6d9b01cb435803cfede1b7dfb782e74f9343ec27fc7 
- Navigate the webacht using keyboard and screen reader and see if the success criteria is satisfied

### DIFF
--- a/cypress/integration/messages/buttons.spec.ts
+++ b/cypress/integration/messages/buttons.spec.ts
@@ -53,4 +53,14 @@ describe("Message with Buttons", () => {
             
         });
     })
+
+    it("buttons in group should have 'aria-label' attribute with button position and name", () => {
+        cy.withMessageFixture('buttons', () => {
+            cy.wrap([1,2,3,4]).each(number => {
+                cy.contains(`foobar005b${number}`)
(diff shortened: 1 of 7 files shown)


pedily:
@sushmi21 to avoid regressions, I'd suggest adding tests here. You can load message fixtures and then expect them to have the "x out of y" aria-labels and identifiers.

sushmi21:
> @sushmi21 to avoid regressions, I'd suggest adding tests here. You can load message fixtures and then expect them to have the "x out of y" aria-labels and identifiers.

Fixed

pedily (approved):

PR #271, opened 2022-05-12
Improve keyboard a11y of time fields in datetime pickers

sushmi21:
Improved the accessibility of DateTime pickers.

Success Criteria:
- The DateTime Pickers should be keyboard accessible

How to test:
- You can use this endpoint to test this PR: https://endpoint-dev-v4.cognigy.ai/599d6418ab91a6f455920996c53347e2a13d660ea1652707ab77bb0ff469a41b
- Test all the Date pickers and DateTime pickers by navigating them using 'TAB' and 'SHIFT +TAB' keys
Previously, you will not be able to navigate out of the time input field (keyboard trap). This should be fixed now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4722,9 +4722,8 @@
     },
     "node_modules/cypress-real-events": {
       "version": "1.7.0",
-      "resolved": "https://cognigy.pkgs.visualstudio.com/_packaging/cognigy-feed/npm/registry/cypress-real-events/-/cypress-real-events-1.7.0.tgz",
-      "integrity": "sha1-rWp43jOvOvDmQ39ccT4waRxERyw=",
-      "license": "MIT",
+      "resolved": "https://registry.npmjs.org/cypress-real-events/-/cypress-real-events-1.7.0.tgz",
+      "integrity": "sha512-iyXp07j0V9sG3YClVDcvHN2DAQDgr+EjTID82uWDw6OZBlU3pXEBqTMNYqroz3bxlb0k+F74U81aZwzMNaKyew==",
       "peerDependencies": {
         "cypress": "^4.x || ^5.x || ^6.x || ^7.x || ^8.x || ^9.x"
       }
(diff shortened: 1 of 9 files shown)


pedily:
Let's add tests for this as well.
I found an article that describes how we could test keyboard navigation and its outcome with cypress:
https://glebbahmutov.com/blog/test-app-using-the-keyboard/#deleting-the-first-item

sushmi21:
> Let's add tests for this as well. I found an article that describes how we could test keyboard navigation and its outcome with cypress: https://glebbahmutov.com/blog/test-app-using-the-keyboard/#deleting-the-first-item

Very useful library for a11y testing. Thanks for suggesting it. :)

sushmi21:
I am not very sure why test runs fail saying 'Unable to authenticate'.  I don't have issues running the tests locally

package-lock.json
```diff
@@ -4763,6 +4720,15 @@
         "node": ">=12.0.0"
       }
     },
+    "node_modules/cypress-real-events": {
+      "version": "1.7.0",
+      "resolved": "https://cognigy.pkgs.visualstudio.com/_packaging/cognigy-feed/npm/registry/cypress-real-events/-/cypress-real-events-1.7.0.tgz",
```
pedily, line 4725: Please reinstall this without your `.npmrc` redirecting it to a private repo

sushmi21: Fixed. thanks!

pedily:
@sushmi21 I created a separate request item to handle the "access to the webchat root element" now. It's recurring in many places and it makes our code more vulnerable to inconsistencies.

If a customer has a `<div id="webchatWindow"></div>` on their page, this _breaks_ our webchat.

sushmi21:
> @sushmi21 I created a separate request item to handle the "access to the webchat root element" now. It's recurring in many places and it makes our code more vulnerable to inconsistencies.
> 
> If a customer has another `<div id="webchatWindow"></div>` on their page, this _breaks_ our webchat when we use `document.querySelector`. Instead, having access to a ref of the webchat root element would be safer. `rootElement.querySelector` would take away the risk of "searching in the wrong place"

Since you have created a new ticket, should we just merge this? Or do you want me to handle it in this PR?

PR #277, opened 2022-08-10
Update datepicker, add possibility to disable or enable dates by fn

Catharsis68:
This PR adds the possibility to disable or enable dates by function.

* This functions was executed before in service-ai in a VM to ensure that the function does not contain any malformed code.
* The function is forced to return a boolen or null
* Added a cypress test where each date is alternately deactivated. Starting with the next | previous day.

### DIFF
--- a/cypress/fixtures/messages/date-picker-function-invalid.json
+++ b/cypress/fixtures/messages/date-picker-function-invalid.json
@@ -1,0 +1,24 @@
+{
+	"text": "",
+	"data": {
+		"_plugin": {
+			"type": "date-picker",
+			"data": {
+				"eventName": "foobar012",
+				"locale": "en",
+				"enableTime": false,
+				"mode": "single",
+				"wantDisable": true,
+				"enable_disable": null,
+				"function_enable_disable": "(date) => { return date.invalidFunction() }",
+				"minDate": "",
+				"maxDate": "",
+				"openPickerButtonText": "foobar012b1",
+				"cancelButtonText": "foobar012b2",
+				"submitButtonText": "foobar012b3",
+				"time_24hr": true,
+				"dateFormat": ""
+			}
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


EgleHelms:
@Catharsis68 generally works well, but functions invalid return statement seem to crash the webchat. I think we need to prevent that with maybe  just showing the calendar with all dates active or so.

Catharsis68:
@EgleHelms  Thank you for reviewing. 
I'm not sure if we'd need this. As mentioned above we already execute the function code in service ai and only if the fn is valid (returns a boolean) send it to the webchat widget. Otherwise the property is an empty string.

EgleHelms (approved):

cbuchwald:
For me the date picker test fails with a timeout

EgleHelms (changes requested):
@Catharsis68 I asked Robin for his opinion and unfortunately, we still came to the same conclusion, that the webchat still needs to be able to handle invalid functions without crashing, e.g. just ignore that particular feature and treat it as if it wasn't set, maybe log a warning. Since webchat has capabilities with sending a message when started with a method etc, that customers may use.

Catharsis68:
@EgleHelms Invalid flatpickr functions are now getting catched. Added also a test case for this scenario.

EgleHelms:
@Catharsis68 it fails for me now too.
image

Anurag9589 (approved):
Looks good to me.

invalid function, error, typo, do not crash the webchat. There is only a true or false "boolean" return function that works, any exception does not crash the webchat.

All specs are passing.
image

EgleHelms (approved):

EgleHelms:
@Catharsis68 again the same thing when the date shifted:
image

Catharsis68:
Fixed, I was calculating the have/not have class on the day of a date (e.g 2022-08-26 > 26)  instead of the weekday > 5 (friday).

EgleHelms (approved):

PR #282, opened 2022-10-12
Add "Inner Socket Handshake"

pedily:
...by upgrading @cognigy/socket-client to v4.6.3.

If the `innerSocketHandshake` option is set in `initWebchat`, the connection should be established in a way that `userId`, `sessionId` and `URLToken` are not passed through URL params, but through the websocket after the connection.

### DIFF
--- a/docs/embedding.md
+++ b/docs/embedding.md
@@ -50,17 +50,18 @@ You can pass [Webchat Options](#webchat-options) as an additional argument to th
 
 ### Webchat Options
 
-| Name              | Type                                    | Default                                          | Description                                                                                                      |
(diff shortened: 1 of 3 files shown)


dshire:
Not sure why this doesn't seem to work for me. Initiated a local webchat with enableInnerSocketHandshake: true, connected to dev-v4. We can maybe have a look together

dshire (approved):
works as expected

dshire:
> Not sure why this doesn't seem to work for me. Initiated a local webchat with enableInnerSocketHandshake: true, connected to dev-v4. We can maybe have a look together

enableInnerSocketHandshake: true must be set on top-level, not in the settings object

PR #283, opened 2022-11-09
Feature/32996 add download button for image attachment

sushmi21:
Add download button for image attachment. 

- Run the new test 'downloadableImage' and check if the download button is shown below the image.

### DIFF
--- a/cypress/fixtures/messages/downloadableImage.json
+++ b/cypress/fixtures/messages/downloadableImage.json
@@ -1,0 +1,39 @@
+{
+	"text": null,
+	"data": {
+		"_cognigy": {
+			"_default": {
+				"_image": {
+					"type": "image",
+					"imageUrl": "https://placekitten.com/300/300",
+					"buttons": [
+						{
+							"id": 0.17194659386513078,
+							"payload": "",
+							"type": "web_url",
+							"title": "Download",
+							"url": "https://placekitten.com/300/300"
+						}
+					]
+				}
+			},
+			"_webchat": {
+				"message": {
+					"attachment": {
+						"type": "image",
+						"payload": {
+							"url": "https://placekitten.com/300/300",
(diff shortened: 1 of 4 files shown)


cypress/fixtures/messages/downloadableImage.json
```diff
@@ -0,0 +1,39 @@
...
+    "data": {
+        "_cognigy": {
+            "_default": {
+                "_image": {
+                    "type": "image",
+                    "imageUrl": "https://placekitten.com/300/300",
+                    "buttons": [
+                        {
+                            "id": 0.17194659386513078,
+                            "payload": "",
```
dshire, line 12: the id and payload fields are not used or are they? I didn't see them in your other PR. are they needed for the test?

src/plugins/messenger/MessengerPreview/components/MessengerMediaTemplate/MessengerMediaTemplate.tsx
```diff
@@ -67,8 +80,26 @@ export const getMessengerMediaTemplate = ({
             }, 100);
         }
 
-        const { media_type, url, altText } = element as IFBMMediaTemplateUrlElement;
-        // TODO add buttons
+        const imageButtons = () => {
```
dshire, line 83: should we not return null here if buttons is falsy? else you would return an empty div

cypress/fixtures/messages/downloadableImage.json
sushmi21: Yes, they were added for tests. I first wrote a test that uses the id. But I will remove this as it is not used anymore

sushmi21: removed

dshire (approved):

PR #285, opened 2022-11-18
Feature/32710 add option to render all urls in a text message as html link

dshire:
#### This PR adds the option to render all urls in a text message as html links.
- works for regular messages from any source
- also works for message with quick replies (the other message types have their own way to add urls)
- default is active
- you can disable it with the new settings option `disableRenderURLsAsLinks: true`

#### Some example strings you can test:
https://www.google.com
http://google.com
go to google.com for this
://www.google.com
find it at www.google.com/test?number=123#end or http://www.bing.com please
<a target=”_blank” href="http://www.cognigy.com">Cognigy Homepage</a>


cbuchwald:
Unfortunately it breaks my urls. I created a say node with "go to https://google.com to test" and it converts the link target to an anchor element <a target="blank" href="//https://google.com">https://google.com</a> which gets resolved in my browser (Chrome 107) as http//google.com missing the colon

dshire:
> Unfortunately it breaks my urls. I created a say node with "go to https://google.com to test" and it converts the link target to an anchor element `<a target="blank" href="//https://google.com">https://google.com</a>` which gets resolved in my browser (Chrome 107) as http//google.com missing the colon

fixed, thanks

sushmi21:
Would be nice to have a couple of tests for this.

src/plugins/messenger/MessengerPreview/components/MessengerTextWithQuickReplies/MessengerTextWithQuickReplies.tsx
```diff
@@ -79,7 +80,9 @@ export const getMessengerTextWithQuickReplies = ({
             setTimeout(() => {quickReplyButton?.focus()}, 200);
         }, []);
 
-        const __html = config.settings.disableHtmlContentSanitization ? text : sanitizeHTML(text);
+		const enhancedURLsText = config.settings.disableRenderURLsAsLinks ? text : replaceUrlsWithHTMLanchorElem(text);
+
+		const __html = config.settings.disableHtmlContentSanitization ? enhancedURLsText : sanitizeHTML(enhancedURLsText);
```
sushmi21, line 85: Formatting is off.

sushmi21:
The color of the link from quick replies is not white. I remember we fixed the color of the links in the past. Was it not? The link in the second bubble is from Quick replies and the first one is from the Say node.
Screenshot from 2022-11-21 10-15-05

src/plugins/messenger/MessengerPreview/components/MessengerTextWithQuickReplies/MessengerTextWithQuickReplies.tsx
dshire: fixed

dshire:
> The color of the link from quick replies is not white. I remember we fixed the color of the links in the past. Was it not? The link in the second bubble is from Quick replies and the first one is from the text output. Screenshot from 2022-11-21 10-15-05

Good point, but let's ignore this for now as it's only related but not blocking the user story

cbuchwald (approved):

sushmi21 (approved):

PR #288, opened 2022-11-29
Bug/14769 handle phone number button type

sushmi21:
Success Criteria
- Clicking on a Phone number button should open the Phone app to place a call
- Phone number should rendered using an anchor element instead of button element
- Link is rendered without issues even if Phone number payload is missing
- Anchor element used to render phone number should look similar to the rest of the button elements
- The button elements should remain unchanged and work as before

How to test:
- Create a Flow and add a Say node of output type text with buttons
- Add a button of type phone number and enter the phone number as payload
- Create a webchat endpoint and use this to test the above success criteria

### DIFF
--- a/src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
+++ b/src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
@@ -50,7 +50,7 @@ export const getMessengerButton = ({ React, styled }: MessagePluginFactoryProps)
         const buttonType = button.type;
         const buttonPayload = button.payload;
         const isWebURL = buttonType === "web_url";
-        const isPhoneNumber = buttonType === "phone_number";
+        const isPhoneNumber = buttonPayload && buttonType === "phone_number";
         const buttonTitle = button.title ? button.title + ". " : "";
         const isWebURLButtonTargetBlank = button.target !== "_self";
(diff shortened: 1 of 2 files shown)


src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
```diff
@@ -14,33 +14,43 @@ interface IMessengerButtonProps {
 
 export const getMessengerButton = ({ React, styled }: MessagePluginFactoryProps) => {
 
-    const Button = styled.button(({ theme }) => ({
+    const ComponentStyles = (({ theme }) => ({
```
dshire, line 17: ComponentStyles is not an element, should start with lowercase letter

dshire (changes requested):
There is also a "Phone Number" option for quick replies in "Tex with Quick Replies" say node variant. I think you missed that one. It only posts back the phone number as user input when I now click it

src/plugins/messenger/MessengerPreview/components/MessengerButton/MessengerButton.tsx
sushmi21: Fixed

sushmi21:
> There is also a "Phone Number" option for quick replies in "Tex with Quick Replies" say node variant. I think you missed that one. It only posts back the phone number as user input when I now click it

As discussed, I will create a separate ticket for this.

dshire (approved):

PR #296, opened 2023-01-25
Polish File attachment UI

sushmi21:
- Set enableFileAttachment flag to true to test this PR
- When clicking on the attchment icon next to the send message button, file explorer should open allowing you to select files to attach
- Alternatively, when you drag a file to the bottom of the webchat, a dropzone should appear. Once you drop your file in the dropzone, the dropzone should disappear and you should be able to see the list of attached files below your text input field.
- The drag and drop behavior should be smooth
- You should still be able to remove a file from the attachment like before


dshire (approved):
lgtm

when did we change this setting from enableFileInput to enableFileAttachment?

PR #298, opened 2023-02-10
Bug/27520 adaptive cards not scrolling

fabclj:
**This PR fix the scrollTo bug for Adaptive Cards and adds the option to scroll to a specific position. In case of Adaptive Card plugin it is the offsetTop.**

- After rendering a message of type Adaptive Card the scrollTo functionality should work as expected.
- The ChatScroller Component now accepts a optional callback called "setScrollToPosition" and a numeric value. If both are defined it scrolls to the defined value and then reset to 0, otherwise scroll to bottom by default.
- The AdaptiveCard plugin component now accepts a optional callback called "setCardOffsetTop". If the callback is defined, it's invoked passing the offsetTop value of card for custom usage (ex.: global state). This can be useful if the adaptive card is expected to be higher then the clientHeight, and is preferable to scroll to the top of the card.

**How to test:**

- Create a Flow and add a Say node of output type Adaptive Card.
- Create a webchat endpoint and use this to test the above success criteria.

### DIFF
--- a/src/webchat-ui/components/history/ChatScroller.tsx
+++ b/src/webchat-ui/components/history/ChatScroller.tsx
@@ -7,8 +7,8 @@ const CLIENT_HEIGHT_OFFSET = 16 + 70; // banner + typing indicator
 export interface OuterProps extends React.HTMLProps<HTMLDivElement> {
     disableBranding: boolean;
     showFocusOutline: boolean;
-    forceScrollingTo: number;
-    setForceScrollingTo?: (offsetTop: number) => void;
+    scrollToPosition: number;
+    setScrollToPosition?: (position: number) => void;
     tabIndex: 0 | -1;
  }
 
@@ -91,10 +91,10 @@ export class ChatScroller extends React.Component<InnerProps, IState> {
     }
 
     componentDidUpdate(prevProps: InnerProps, prevState: IState, wasScrolledToBottom: boolean) {
-        const { setForceScrollingTo, forceScrollingTo } = this.props;
-        if(forceScrollingTo && setForceScrollingTo) {
-            this.scrollToPosition(forceScrollingTo);
-            setForceScrollingTo(0);
+        const { setScrollToPosition, scrollToPosition } = this.props;
+        if(scrollToPosition && setScrollToPosition) {
+            this.scrollToPosition(scrollToPosition);
+            setScrollToPosition(0);
         } else if (wasScrolledToBottom) {
             this.scrollToBottom();
         }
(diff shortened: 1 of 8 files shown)


src/webchat-ui/components/history/ChatScroller.tsx
```diff
@@ -73,7 +91,11 @@ export class ChatScroller extends React.Component<InnerProps, IState> {
     }
 
     componentDidUpdate(prevProps: InnerProps, prevState: IState, wasScrolledToBottom: boolean) {
-        if (wasScrolledToBottom) {
+        const { setScrollToPosition, scrollToPosition } = this.props;
+        if(scrollToPosition && setScrollToPosition) {
+            this.scrollToPosition(scrollToPosition);
```
dshire, line 96: naming is a bit weird here. maybe change the name for the function to make sure that these two are not the same (not even same type). Examples:  onScrollToPosition or handleScrollToPosition

> this.scrollToPosition(scrollToPosition);

src/webchat-ui/components/history/ChatScroller.tsx
```diff
@@ -62,6 +64,22 @@ export class ChatScroller extends React.Component<InnerProps, IState> {
...
+        // we need the container reference to perform a scroll on it
+        if (!root)
+            return;
+
+        try {
+            root.scroll({
+                top: position - 70
+            });
+        } catch (e) {
+            root.scrollTop = position;
```
dshire, line 79: why this error handling?

>     root.scrollTop = position;

dshire (changes requested):
looks good, just a naming related comment and a question

src/webchat-ui/components/history/ChatScroller.tsx
fabclj: Yes I agree. I think I could merge the two methods "scrollToBottom" and "scrollToPosition" in only one called "handleScrollTo", accepting an optional number parameter. If it is defined we scroll to that value, otherwise we follow the default logic to scroll to the bottom.

src/webchat-ui/components/history/ChatScroller.tsx
fabclj: I just used the same idea from the existing method "scrollToBottom". If the instance method "scroll" fails we do scrolling by setting "scrollTo" instance parameter.
Anyway, if we merge the two methods as suggested in the above comment, this error handling will not change.

dshire:
I just noticed another thing @fabclj : In my bot I mentioned above in my comment, you get a message and then an adaptive card. I think that is a common use case, imagine a call to action (Please choose a new seat!) and then the card (a widget to select a seat for your flight). We also have to see the message. So it is actually more complex, we have to scroll to a position where all bot messages that came in - after the last user input - before the adaptive card - are also in scroll view.

Do you agree? Maybe we need to discuss this further with PO, but I don't want to fix this into another breaking behavior (current behavior you would see the message, so merging like this would break/fix it)

dshire (approved):

PR #299, opened 2023-02-14
Bug/37877 typeguard the url recognition against wrong types of text

dshire:
https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/37877

Webchat can break if you send a bot output with a different type than text.

To recreate the issue, connect to a bot that has a code node. In the code node, write something like api.say(123) (or send an object etc.) Then connect to a webchat endpoint and talk to the bot through webchat.


sushmi21 (approved):
Works as expected

PR #300, opened 2023-02-20
Feature/37385 image lightbox

fabclj:
This PR implements the possibility to left-click on one image preview and open a lightbox/modal containing the image in full size, responsive. On top right there is the button for close the modal and download the original file.

**How to test:**
Create a inbox conversation with LA and use it to send a message with image.


src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,73 @@
...
+            if (!isSameDomain) {
+                window.open(url, '_blank');
+            }
+        }
+
+        return (
+            <Header>
+                <Caption>{altText}</Caption>
+                <IconsGroup>
+                    <Icon onClick={handleDownload} download>
```
sushmi21, line 61: Would be nice to make this button accessible by 
1. Adding visible focus style
2. Adding meaningful aria-label

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        <DownloadIcon />
+                    </Icon>
+                    <Icon onClick={onClose}>
```
sushmi21, line 64: Would be better to use button element instead of anchor element and style it similar to the download button. Else, use role="button".

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+            if (!isSameDomain) {
+                window.open(url, '_blank');
+            }
+        }
+
+        return (
+            <Header>
+                <Caption>{altText}</Caption>
+                <IconsGroup>
+                    <Icon role="button" onClick={handleDownload} download aria-label="Download fullsize image">
```
sushmi21, line 64: The buttons still don't receive keyboard focus. We need to add tabIndex=0

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+import DownloadIcon from '../../../../../../webchat-ui/assets/baseline-download-24px.svg';
+
+export const getLightBoxHeader = ({ React, styled }: MessagePluginFactoryProps) => {
+
+    const Header = styled.div({
+        position: 'absolute',
+        top: 0,
+        height: '56px',
+        width: '100%',
+        backgroundColor: 'rgba(0, 0, 0, 0.7)',
```
sushmi21, line 13: i think it would be nice to use the primary color from the theme and to style the buttons in the header in the same way as that of the webchat header

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    const ImagePreview = () => {
+        const { config, url, altText, template, onExpand } = useMessangerImageContext();
+
+        const handleKeyDown = (event: KeyboardEvent) => {
+            event.code === "Enter" && onExpand && onExpand();
+        }
+
+        return config.settings.dynamicImageAspectRatio ? (
+                <FlexImage
```
sushmi21, line 41: We can improve the focus style of these images and make it consistent with the rest of the elements. We can remove the outline and add an inset box shadow that uses the primary weak color from the theme (For example. 'box-shadow: inset 0 0 0px 1px <theme.primaryWeakColor>'. Similar to focusing audio and video elements)

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+                    data-test="image-preview"
+                    tabIndex="0"
+                    onKeyDown={handleKeyDown}
+                />
+            ) : (
+                <FixedImage
+                    style={{ backgroundImage: getBackgroundImage(url) }}
+                    onClick={() => onExpand()}
+                    template={template}
+                    data-test="image-preview"
```
sushmi21, line 54: Now that we add tabIndex=0, we must add role="button" and an aria-label to this wrapper. Please do not add the button role and aria-label to FlexImage component, since it uses the image element. We can instead wrap the FlexImage in a div that acts as a button. We need to be sure that adding this div does not break anything.

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,97 @@
+import { useMessangerImageContext } from '../MessangerImageContext';
+import { MessagePluginFactoryProps } from "../../../../../../common/interfaces/message-plugin";
+import CloseIcon from '../../../../../../webchat-ui/assets/baseline-close-24px.svg';
+import DownloadIcon from '../../../../../../webchat-ui/assets/baseline-download-24px.svg';
+
+export const getLightBoxHeader = ({ React, styled }: MessagePluginFactoryProps) => {
```
sushmi21, line 6: When I open the lightbox and tab through it, only the download and close buttons should receive keyboard focus (That is, focus should be trapped within the lightbox). But all the interactive elements in the webchat are receiving focus, which should not be the case.

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
fabclj: right now I just added some opacity on hover and focus. Can be improved later on next commit

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+        margin: 0,
+        transition: 'background-color .1s ease-out,color .1s ease-out,fill .1s ease-out',
+        color: 'white',
+        borderRadius: '50%',
+        cursor: 'pointer',
+        "& svg": {
+            fill: 'white',
+        },
+        "&:hover, &:focus": {
+            backgroundColor: 'hsl(211,60%,33%)',
```
sushmi21, line 50: We can use the primaryStrongColor from the theme

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+        overflow: 'hidden',
+        display: 'flex',
+        justifyContent: 'space-between'
+    }))
+
+    const Caption = styled.div({
+        display: 'flex',
+        alignItems: 'center',
+        marginLeft: '15px',
+        color: 'white',
```
sushmi21, line 24: We should use primaryContrastColor from the theme, as the text will be hard to read on light backgrounds (White or other light colors might also be the primary color).

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    const Icon = styled.button({
+        padding: '8px',
+        boxSizing: 'border-box',
+        backgroundColor: 'transparent',
+        border: 'none',
+        outline: 'none',
+        margin: 0,
+        transition: 'background-color .1s ease-out,color .1s ease-out,fill .1s ease-out',
+        color: 'white',
```
sushmi21, line 43: Same here, we should use primaryContrastColor from these, so that it is visible on white backgrounds. Also same for 'fill' in line 47.

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,73 @@
...
+            event.code === "Enter" && onExpand && onExpand();
+        }
+
+        return config.settings.dynamicImageAspectRatio ? (
+                <div tabIndex={0} role="button" aria-label="View Image in fullsize" onClick={() => onExpand()}>
+                    <FlexImage
+                    src={url}
+                    alt={altText || "Attachment"}
+                    data-test="image-preview"
+                    onKeyDown={handleKeyDown}
```
sushmi21, line 53: The key down handler should also be moved to the div

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    const FixedImage = styled.div({
+        backgroundSize: "cover",
+        backgroundPosition: "center center",
+        cursor: 'pointer',
+        "&:hover": {
+            opacity: '.8'
+        },
+        "&:focus": {
+            opacity: '.8',
```
sushmi21, line 29: I tested the focus style, I think we can leave it with the opacity instead of border or box-shadow. But also please add the same focus style for FlexImage

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,85 @@
...
+        display: 'block',
+        width: '100%',
+        cursor: 'pointer',
+        "&:hover": {
+            opacity: '.8'
+        },
+        "&:focus": {
+            opacity: '.8',
+            outline: 'none'
+        },
```
sushmi21, line 44: This won't work. The focus style should be added to the div wrapping this with tabIndex=0

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/LightBoxHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+        return (
+            <Header>
+                <Caption>{altText}</Caption>
+                <IconsGroup>
+                    <Icon
+                        ref={firstButton}
+                        onClick={handleDownload}
+                        onKeyDown={handleKeyDownload}
+                        download
+                        tabIndex="0"
```
sushmi21, line 88: tabIndex is not needed anymore since we changed the Icons to be a button. Same for the close button

sushmi21 (approved):
Let us not merge this and wait for approval from Thomas regarding the expansion of images from Say node. Rest looks really good.

src/plugins/messenger/MessengerPreview/components/MessengerImage/components/ImagePreview.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const FixedImage = styled(FlexImageDiv)({
+        backgroundSize: "cover",
+        backgroundPosition: "center center",
+    }, templateStyle);
+
+    const FlexImage = styled.img({ 
+        display: 'block',
+        width: '100%',
+        cursor: 'pointer'
```
sushmi21, line 40: We need to remove the pointer for FlexImage if the image is not downloadable

sushmi21 (approved):

PR #301, opened 2023-02-24
Fix color for thumbs up/down icon in header

sushmi21:
- Add enableRating: "always" setting to see the rating icon in the webchat header
- Configure your endpoint, such that it uses a light color as primary color
- The rating icon should now use the contrast color, instead of white color


fabclj (approved):
looks good

PR #303, opened 2023-03-09
Add class names to rating dialog elements

sushmi21:
Add class names to rating dialog elements to make it more customizable by the user.

Success Criteria:
- All essential elements in the Rating dialog now have a class name
- Added classnames are documented in the mark down file


dshire (approved):
lgtm

PR #327, opened 2023-08-14
Bug/48743 Minimizing issues with adaptive cards

fabclj:
**This PR fix the scrollToPosition bug for Adaptive Cards on minimizing the chat or on first render**

- After rendering a history chat with at least a message of type Adaptive Card, the chat should scroll to bottom, not to last Adaptive Card message.
- On minimizing a chat and expand again, the chat should scroll to bottom, not to last Adaptive Card message.

**How to test:**

- Create a Flow and add a Say node of output type Adaptive Card.
- Add some say node of type Text after the Adaptive Card
- Create a webchat endpoint and use this to test the above success criteria.

**NOTE**: It's important to test for new/empty chat session, and also with long chat session. Testing also page refresh and minimizing action.

### DIFF
--- a/src/webchat-ui/components/WebchatUI.tsx
+++ b/src/webchat-ui/components/WebchatUI.tsx
@@ -62,7 +62,9 @@ export interface WebchatUIProps {
     onConnect: () => void;
     onToggle: () => void;
     onSetScrollToPosition: (position: number) => void;
+    onSetLastScrolledPosition: (position: number | null) => void;
     scrollToPosition: number;
+    lastScrolledPosition: number | null;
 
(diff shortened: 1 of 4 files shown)


fabclj:
@sushmi21 I also noticed some issue like that you described, but only with large chat history. Since we are not using virtualized list, in case of a big amount of messages, the browser can lead to performance problem.
If having such a large amount of message is a common user case, we really need to consider to switch the chat list to a virtualized list. I would consider this after talking to PO and in a new PR, since the implementation could require relevant effort.

sushmi21 (approved):

PR #338, opened 2023-11-15
Feature/56144 home screen

dshire:
Added basic home screen, colors, some new components and cleaned up some older components

task https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/56144
part of US https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/55708


src/webchat-ui/components/presentational/Button.tsx
```diff
@@ -12,46 +13,52 @@ export default styled.button<IColorProps>(({ color, theme }) => {
     }
 
     if (color === 'primary') {
```
fabclj, line 15: Here I don't get why we set secondaryColor if when color is equal to 'primary'

fabclj:
the buttons have 16px font-size, but the expected should be 14px
Screenshot from 2023-11-16 10-10-51

src/webchat-ui/components/presentational/Button.tsx
dshire: According to the design specs, the so called "primary button" is using the so called "secondary colors"

dshire:
> the buttons have 16px font-size, but the expected should be 14px Screenshot from 2023-11-16 10-10-51

fixed

fabclj (approved):

PR #342, opened 2023-11-30
implement notification banner

kwinto:
This PR adds notification banner implementation according to Azure: #55714. There is currently no use of it. 

We expose it as `createNotification` method to be used later by us but also for the customers we expose `webchat.showNotification()` API method.

# How to test
Add this code after :26 in index.html
```
webchat.open();
setTimeout(() => webchat.showNotification("Test notification"), 1000);
```


kwinto:
@fabclj @dshire I've removed test run for the ES legacy support check as we are dropping legacy browser support.

If you believe there is a reason to keep, just ping me 

https://github.com/Cognigy/WebchatWidget/pull/342/commits/8d019b3d34e9c4959d710667342ae278c0dacccf

PR #343, opened 2023-12-01
Previous conversations overview

fabclj:
This PR includes the first part of a bigger task about Previous conversations screens.
It was completed the Overview Screen and the Redux reducer setup that can be continued on next PR.

reference US #355723

### DIFF
--- a/src/webchat/helper/storage.ts
+++ b/src/webchat/helper/storage.ts
@@ -23,29 +23,30 @@ export const getAllConversationsByUserID = (
 	currentUserId?: string,
 	currentSessionId?: string,
 ) => {
-	return Object.keys(storage).reduce((obj, k) => {
+	return Object.keys(storage).reduce((acc, item) => {
 		// skip missing userId or sessionId
-		if (!currentSessionId || !currentUserId) return obj;
+		if (!currentSessionId || !currentUserId) return acc;
 
-		const data = storage.getItem(k) || "";
+		const data = storage.getItem(item) || "";
 
 		// skip if not JSON
-		if (!isValidJSON(k) || !isValidJSON(data)) return obj;
+		if (!isValidJSON(item) || !isValidJSON(data)) return acc;
 
-		const sessionId = JSON.parse(k)?.[2] || "";
-		const userId = JSON.parse(k)?.[1] || "";
+		const sessionId = JSON.parse(item)?.[2] || "";
+		const userId = JSON.parse(item)?.[1] || "";
 
 		// skip different userID
-		if (currentUserId !== userId) return obj;
+		if (currentUserId !== userId) return acc;
 
 		// skip current sessionId
-		if (sessionId === currentSessionId) return obj;
+		if (sessionId === currentSessionId) return acc;
 
 		const messages = JSON.parse(data)?.messages;
 
 		// skip empty
-		if (!Array.isArray(messages) || messages.length === 0) return obj;
-
-		return { ...obj, [sessionId]: JSON.parse(data) };
+        if (!Array.isArray(messages) || messages.length === 0) return acc;
+        
+        acc[sessionId] = JSON.parse(data);
+        return acc;
 	}, {});
 };


kwinto:
Missing sender title and avatar placeholder (or a shift) when last message is from the user
Screenshot 2023-12-01 at 12 44 04

kwinto:
It should truncate text, currently is wrapping the line (Safari)
Screenshot 2023-12-01 at 12 48 19

src/webchat/helper/storage.ts
```diff
@@ -1,18 +1,51 @@
...
+
+		// skip current sessionId
+		if (sessionId === currentSessionId) return obj;
+
+		const messages = JSON.parse(data)?.messages;
+
+		// skip empty
+		if (!Array.isArray(messages) || messages.length === 0) return obj;
+
+		return { ...obj, [sessionId]: JSON.parse(data) };
```
kwinto, line 49: It's better to assign a new property to the accumulator object and return it instead of creating a new object every time:
```javascript
obj[sessionId] = JSON.parse(data);

return obj
```

src/webchat/helper/message.ts
```diff
@@ -1,19 +1,28 @@
-import { IMessage } from "../../common/interfaces/message"
+import { IMessage } from "../../common/interfaces/message";
 
 const getTextFromMessage = (message: IMessage) => {
+	// Check if message is plain text
+	if (message.text) {
+		return message.text;
+		// Check if message is quick reply message
+	} else if (message?.data?._cognigy?._webchat?.message?.text) {
```
kwinto, line 8: Here and down you want to use `getMessengerPayload` function and not access the `_webchat` directly as a message can come from a different channel (_facebook or _defaultPreview)

src/webchat-ui/components/presentational/previous-conversations/helpers.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const getParticipants = (messages: IMessage[], config: IWebchatConfig) => {
+	const partecipants: string[] = [];
+	const hasBot = messages.some(message => ["bot", "engagement"].includes(message?.source));
+	const hasLiveAgent = messages.some(message => message?.source === "agent");
+
+	// TODO: get the correct names, if any, from the endpoint
+
+	if (hasBot) partecipants.push(config?.settings?.title || "Bot");
+	if (hasLiveAgent) partecipants.push("Live Agent");
```
kwinto, line 77: Just "Agent"

src/webchat-ui/components/presentational/previous-conversations/helpers.ts
```diff
@@ -0,0 +1,104 @@
...
+		case "template":
+			return "template...";
+		case "image":
+			return "image...";
+		case "video":
+			return "video...";
+		case "audio":
+			return "audio...";
+		default:
+			return "...";
```
kwinto, line 65: I would use some general text such as "Message" for the default. Ellipsis can be confused with "currently typing" UI convention.

Also capitalize all the items: Image, Video, ...

src/webchat-ui/components/presentational/previous-conversations/OverviewScreen.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+	const sessions = Object.keys(sortedConversations);
+
+	const handleStartButtonClick = () => {
+		onSetShowPrevConversationsScreen(false);
+	};
+
+	return (
+		<ScreenRoot className="webchat-homescreen-root">
+			<ScreenContent className="webchat-homescreen-content">
```
kwinto, line 78: Are "webchat-homescreen..." classes intended here?

src/webchat-ui/components/presentational/HomeScreen.tsx
```diff
@@ -101,18 +95,27 @@ interface IHomeScreenProps {
...
+		onSetShowPrevConversationsScreen(true);
+	};
+
 	return (
 		<HomeScreenRoot className="webchat-homescreen-root">
 			<HomeScreenContent className="webchat-homescreen-content">
 				<HomeScreenHeader className="webchat-homescreen-header">
+					<div className="webchat-homescreen-header-title">
+						{config?.settings?.title || "Cognigy Webchat"}
+					</div>
```
kwinto, line 118: A duplicate here

kwinto (changes requested):

src/webchat/helper/message.ts
fabclj: this is a existing function I didn't change and I would not change in PR.

src/webchat-ui/components/presentational/previous-conversations/OverviewScreen.tsx
fabclj: fixed

fabclj:
> Missing sender title and avatar placeholder (or a shift) when last message is from the user 

I fixed this one. To be honest I didn't know this case could be possible.
Solution: In case of user is the only one in the conversation, We will show the placeholder avatar and "You" as Participants string

kwinto:
> Solution: In case of user is the only one in the conversation, We will show the placeholder avatar and "You" as Participants string

It should handle the case where the last message is from the user, even if the conversation has many participants.

fabclj:
> It should handle the case where the last message is from the user, even if the conversation has many participants.

I think we have a different understanding of this feature. On my understanding we show there the participant and this is not depending from the last message source.

src/webchat-ui/components/presentational/previous-conversations/helpers.ts
kwinto: This was not fixed, let's double-check the naming

fabclj: it was fixed here already
https://github.com/Cognigy/WebchatWidget/pull/343/commits/26390c8c85eb290066790f39f4c213f4426401d7

PR #347, opened 2023-12-07
Feature/55708 home screen 2

dshire:
https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/56145

**To review this, please make sure you also review and load the related bundle from chat-components manually**

This PR adds 2 items regarding the home screen:

1. Functional action buttons to the home screen (still hardcoded since there is no endpoint setting yet)
2. The homescreen background is now either calculated from the primaryColor, showing a configured backgroundImage, or showing both at once (image on top of color gradient)

To test the backgroundimage, either initalize with the setting "backgroundImageUrl" set to an image you want to display, or configure this parameter in the endpoint settings (it is the setting for the background image of the demo webchat, we have to adjust this as soon as we have a new endpoint setting).


fabclj:
looks good to me

fabclj (approved):

dshire:
> looks good to me

I'll wait with merging this until we have a new release of the chat components with the exported buttons that we can import here properly

dshire:
> @dshire One small thing to follow up on: Figma design mentions Homescreen QR buttons becoming slightly smaller on mobile.

@kwinto I think that's only for the teaser message buttons? I can't find this description for the homescreen buttons. They are also the same in the desktop & mobile behavior  section

PR #348, opened 2023-12-08
Typing Indicator

kwinto:
This PR adds a newly styled typing indicator from the chat-components repo to the Webchat.


fabclj (approved):
It works as expected

PR #352, opened 2023-12-12
Feature/58356 use typography component

sushmi21:
This PR replaces redundant font styling with Typography component. 

Check if the text styling in the UI is not altered.


fabclj:
It looks good to me.
I am just wondering if we should already plan a second round like that. We have new PRs coming that should also make use of this Typography component

fabclj (approved):

sushmi21:
> It looks good to me. I am just wondering if we should already plan a second round like that. We have new PRs coming that should also make use of this Typography component

Sure. I can do that once the other PRs are merged :)

PR #354, opened 2023-12-13
Fix maintenance mode PR

fabclj:
This PR includes again the changes already done in #351 and removed because conflict solving issues.

**How to test:**
put this as settings:
```
settings: {
    awaitEndpointConfig: true
}
```

Just set maintenance mode or Business hour on target endpoint and check if it appears as expected

### DIFF
--- a/src/webchat-ui/components/WebchatUI.tsx
+++ b/src/webchat-ui/components/WebchatUI.tsx
@@ -788,7 +788,7 @@ export class WebchatUI extends React.PureComponent<
 			<>
 				<Header
 					onClose={handleOnClose}
-					onGoBack={showInformationMessage ? handleOnGoBack : undefined}
+					onGoBack={!showInformationMessage ? handleOnGoBack : undefined}
 					logoUrl={config.settings.headerLogoUrl}
 					title={config.settings.title || "Cognigy"}
 					closeButtonRef={this.closeButtonInHeaderRef}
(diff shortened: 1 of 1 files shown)


src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -755,7 +781,12 @@ export class WebchatUI extends React.PureComponent<
 			);
 
 		const handleOnClose = () => {
-			onClose?.();
+			if (showInformationMessage) {
+				onClose?.();
+			} else {
+				onSetShowPrevConversations(false);
+				onSetShowHomeScreen(true);
+			}
```
kwinto, line 789: Close button should always close the widget. To get back to home screen, there is a back button.

There might be a case where it's not always true (e.g. some pop-up), here though it is not the case.

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -768,14 +799,16 @@ export class WebchatUI extends React.PureComponent<
 			<>
 				<Header
 					onClose={handleOnClose}
-					onGoBack={handleOnGoBack}
+					onGoBack={!showInformationMessage ? handleOnGoBack : undefined}
```
kwinto, line 802: It's better for readability to not have an inversion if possible:
```javascript
onGoBack={showInformationMessage ? undefined : handleOnGoBack}
```

kwinto (changes requested):
some smaller ones, otherwise looks good to me

src/webchat-ui/components/WebchatUI.tsx
fabclj: fixed

src/webchat-ui/components/WebchatUI.tsx
fabclj: fixed

kwinto (approved):

PR #355, opened 2023-12-14
add homescreen button functionality

dshire:
This PR adds the necessary changes to make the homescreen buttons usable. they are still mocked as we don't have the endpoint settings yet, but fully usable as link and postback message.

To test, just deploy a webchat and check the two button types


sushmi21 (approved):
Works as expected

PR #358, opened 2023-12-14
Feature/55721 privacynotice

dshire:
Implements the Privacy Notice feature according to Figma pages and US 55721 (https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/55721)

- adds the visual components to look & behave like the specs
- adds local storage logic so that you only have to accept once
- adds connection logic so that a user only connects to the endpoint after they have accepted
- adds logic that a message action button from home screen is stored and only sent after privacy notice was accepted


sushmi21 (approved):
Looks good

PR #362, opened 2024-01-09
Disable datepicker plugin

fabclj:
This PR disable datepicker plugin since it will be moved to chat-components repo

### DIFF
--- a/src/webchat-embed/index.tsx
+++ b/src/webchat-embed/index.tsx
@@ -6,7 +6,7 @@ import './embedded-webchat-styles.css';
 
 // load plugins
 import '../plugins/get-started-button-input';
-import '../plugins/date-picker';
+// import '../plugins/date-picker';
 import '../plugins/messenger';
 import '../plugins/rating';
 import '../plugins/adaptivecards';


dshire (commented):
should we not fully remove this import line as well as the /date-picker plugin folder? or when do you plan on doing that?

fabclj:
> should we not fully remove this import line as well as the /date-picker plugin folder? or when do you plan on doing that?

Actually we should also remove the build command on package.json, and finally remove all related existing files. I would do these changes in a second step, just after we merge the new datepicker on chat-components and we are sure is properly working on all usecases.

dshire:
> > should we not fully remove this import line as well as the /date-picker plugin folder? or when do you plan on doing that?
> 
> Actually we should also remove any reference on the bundle command on package.json, and finally remove all related existing files. I would do these changes in a second step, just after we merge the new datepicker on chat-components and we are sure is properly working on all usecases.

we can do it like that but then please write a reminder or a task for yourself to do that, else we can forget that we have this unused code

dshire (approved):

fabclj:
> we can do it like that but then please write a reminder or a task for yourself to do that, else we can forget that we have this unused code

I will create a more generic US about plugins and legacy messages code clean-up. I believe we have much more than just datepicker to clean-up on v3 branch.

PR #363, opened 2024-01-09
Feature/59076 add feedback submission in chat history

sushmi21:
This PR adds the 'Feedback submitted' status pill to chat history when submitting rating from 'conversation rating' screen. Additionally, also fixes a bug in Previous Conversation view, where rating screen was shown sometimes instead of the chat history.

Success Criteria:
1. After submitting rating via 'Conversation Rating' screen, a status pill with text 'Feedback submitted' is show in chat history, in addition to the disappearing event banner. 
2. When you open the  'Previous Conversation' view, you should see the chat history and also the status pill within that.
3. The status pill should not appear in chat history anywhere if the rating is submitted via the 'Chat Options' screen

How to test:
- Use an endpoint that points to a flow with a request rating node. You can also use this if you like: https://endpoint-dev.cognigy.ai/a7f6316334debd965e4668ec402043e3498c69f0146fa4b96f29c059337eeab7
- Chat with the bot and provide a rating via the Converation Rating screen. Test SC 1 and 2
- Start another conversation and this time provide rating only via Chat Options screen. Test the SC 3.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			"license": "SEE LICENSE IN LICENSE",
 			"dependencies": {
 				"@braintree/sanitize-url": "^6.0.0",
-				"@cognigy/chat-components": "0.7.0",
+				"@cognigy/chat-components": "0.8.0",
 				"@cognigy/socket-client": "5.0.0-beta.7",
 				"@emotion/cache": "^10.0.29",
 				"@emotion/react": "^11.7.1",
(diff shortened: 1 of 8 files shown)


dshire (approved):
works as expected. 

One thing I noticed, when leaving the chat options or the rating widget, the chat history is loaded new and scrolls down from top to bottom. I feel like we should improve this as it feels rather "buggy". We should probably implement the options and rating (and others) as overlay that just cover the chat instead of rendering it instead of the chat. What do you think?

kwinto:
> works as expected.
> 
> One thing I noticed, when leaving the chat options or the rating widget, the chat history is loaded new and scrolls down from top to bottom. I feel like we should improve this as it feels rather "buggy". We should probably implement the options and rating (and others) as overlay that just cover the chat instead of rendering it instead of the chat. What do you think?

I would just make scroll immediate.

The problem with overlay is you still need to "reserve" space in chat with some calculations.

kwinto (approved):

PR #364, opened 2024-01-11
Feature/56147 home screen animations

dshire:
Added home screen animations like in the figma preview.

To review:
1. deploy webchat
2. go to homescreen
3. open chat or previous conversations (also clear local storage to see privacy note view)
4. go back to home screen to see the inverse animation


sushmi21:
But I noticed small differences between the prototype and the current implementation. 
1. The transition is slightly slower than in Figma.
2. The home-screen contents are not blurred when the transition is in progress. 

Otherwise looks good.

sushmi21 (approved):

dshire:
> But I noticed small differences between the prototype and the current implementation.
> 
> 1. The transition is slightly slower than in Figma.
> 2. The home-screen contents are not blurred when the transition is in progress.
> 
> Otherwise looks good.

1. you are right, the animation in figma is 400ms, i made it 500. I will adjust this now
2. as discussed, for performance reasons I would refrain from blur filters if not specifically requested here

PR #367, opened 2024-01-18
Feature/55722 teaser message

dshire:
Added teaser message according to US https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/55722

To test, make sure that teaser message is enabled. After it was opened, clear local storage and refresh to see it again (Old behavior that stays the same)

NOTE: The Message buttons are still hard-coded and need to be updated with the new v3 settings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			"license": "SEE LICENSE IN LICENSE",
 			"dependencies": {
 				"@braintree/sanitize-url": "^6.0.0",
-				"@cognigy/chat-components": "0.10.0",
+				"@cognigy/chat-components": "0.10.7",
 				"@cognigy/socket-client": "5.0.0-beta.10",
 				"@emotion/cache": "^10.0.29",
 				"@emotion/react": "^11.7.1",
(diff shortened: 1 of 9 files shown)


sushmi21:
Should the action buttons be displayed all the time in the teaser message? I hoped to see them only when a conversation was not already started. But I could see the buttons when I minimize the chat when a conversation is ongoing (See screenshot). Here, I just expected the text message without buttons. Was this already discussed and clarified?
Screenshot from 2024-01-22 09-45-23

dshire:
> Should the action buttons be displayed all the time in the teaser message? I hoped to see them only when a conversation was not already started. But I could see the buttons when I minimize the chat when a conversation is ongoing (See screenshot). Here, I just expected the text message without buttons. Was this already discussed and clarified? Screenshot from 2024-01-22 10-14-57

We don't have any info on the functionality beyond what you see in the specs and what we have in v2. You can see in the User story that I already have some questions to this topics that are yet unanswered. I will add yours as well, but I would not block the PR because of something that we can change later when we know if it needs changes

sushmi21:
> > Should the action buttons be displayed all the time in the teaser message? I hoped to see them only when a conversation was not already started. But I could see the buttons when I minimize the chat when a conversation is ongoing (See screenshot). Here, I just expected the text message without buttons. Was this already discussed and clarified? Screenshot from 2024-01-22 10-14-57
> 
> We don't have any info on the functionality beyond what you see in the specs and what we have in v2. You can see in the User story that I already have some questions to this topics that are yet unanswered. I will add yours as well, but I would not block the PR because of something that we can change later when we know if it needs changes

Okay, I agree. Let's add it as a question for Thomas in the ticket.

sushmi21 (approved):
I will approve the PR, since everything else look good to me.

PR #368, opened 2024-01-18
Added default webchat AI icon to header

dshire:
Added default webchat Cognigy AI icon according to US https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/59582

The variable header size still has the same issues, see the bug ticket

### DIFF
--- a/src/webchat-ui/components/presentational/Header.tsx
+++ b/src/webchat-ui/components/presentational/Header.tsx
@@ -7,6 +7,7 @@ import MenuIcon from "../../assets/menu-16px.svg";
 import Notifications from "./Notifications";
 import classnames from "classnames";
 import { Typography } from "@cognigy/chat-components";
+import CognigyAIAvatar from "../../assets/cognigy-ai-avatar-48px.svg";
 
 const HeaderBar = styled.div(({ theme }) => ({
 	alignItems: "center",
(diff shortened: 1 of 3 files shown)


sushmi21:
Now the chat options header is broken
Screenshot from 2024-01-19 13-34-21

src/webchat-ui/components/presentational/Header.tsx
```diff
@@ -133,8 +135,7 @@ const Header: FC<HeaderProps> = props => {
 
 	const isCompact =
 		!mainContentRef?.current ||
-		mainContentRef.current.scrollHeight > mainContentRef.current.clientHeight ||
-		!logoUrl;
+		mainContentRef.current.scrollHeight > mainContentRef.current.clientHeight;
```
sushmi21, line 138:
```suggestion
		mainContentRef.current.scrollHeight > mainContentRef.current.clientHeight || 
		!isChatOptionsButtonVisible;
```

sushmi21: Since you removed the `!logoUrl `check, we need to add additional condition to make the header always compact in rating and chat options screens.

sushmi21 (changes requested):
I think in the future, we should also make the header logo optional. For now, we could merge this PR if the header is fixed for the chat options screen

dshire:
> I think in the future, we should also make the header logo optional. For now, we could merge this PR if the header is fixed for the chat options screen

Yes that could be another settings option, but then the variable header size looks off and might need changes as well

src/webchat-ui/components/presentational/Header.tsx
dshire: I added your change. I hope we can clear up this mess with the internal navigation soon

sushmi21 (approved):

PR #369, opened 2024-02-02
Rename data attributes for buttons

sushmi21:
Rename data attributes of back and overlay buttons. This PR has no functional impact.


dshire (approved):
lgtm

PR #372, opened 2024-02-06
Teaser Message behavior & style adjustments

dshire:
Some adjustments/fixes to Teaser Message behavior following a product meeting:

1. Teaser message is enabled/disabled dependent on its value being set (`engagementMessageText`)
2. unread message badge is smaller
3. action buttons will not be shown again if the chat was opened once and closed again. a user will see the unread message preview without any buttons (test with `enableUnreadMessagePreview` enabled)


sushmi21:
Can we also change the chat icon as we discussed yesterday? https://mui.com/material-ui/material-icons/?query=cha&selected=ChatBubbleOutline

sushmi21:
The unread message badge is always visible even when there are no unread messages

sushmi21 (approved):

PR #387, opened 2024-03-07
Prevent sending message on IME compositing (v3)

kwinto:
This PR fixes an issue where choosing an input suggestion in, e.g. Japanese IME with enter key, would send the message.

To test you need an MacOS. Add Japanese language, switch to it and start typing into main input field. Press arrow-down key. The suggestions should popup, press enter to choose one of it.

The suggestion should be applied to input, but the message should not be sent to chat.


sushmi21:
One of tests fail in this PR. Can you please check that?

sushmi21 (approved):

PR #393, opened 2024-03-19
Fix multiple issues

kwinto:
This PR fixes following issues:
- Autofocus not working with Galleries
- Clicking a teaser message does not lead to the conversation
- Disconnect Overlay flickering and only visible on HomeScreen (restyled)
[Ticket](https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/62289)

# To test
1. Chat Components: checkout branch, npm pack and install in Webchat [PR](https://github.com/Cognigy/chat-components/pull/40)
2. Add this to the current default settings in index.html
```
widgetSettings: {
	enableAutoFocus: true,
},
behavior: {
	enableConnectionStatusIndicator: true,
},
```

Test for issues to be resolved.

### DIFF
--- a/src/webchat-ui/components/presentational/DisconnectOverlay.tsx
+++ b/src/webchat-ui/components/presentational/DisconnectOverlay.tsx
@@ -95,12 +95,3 @@ const DisconnectOverlay = ({ isPermanent, onClose, onConnect }) => {
 };
 
 export default DisconnectOverlay;
-function alpha(
-	backgroundWebchat: string,
-):
-	| string[]
-	| import("csstype").Property.BackgroundColor
-	| import("csstype").Property.BackgroundColor[]
-	| undefined {
-	throw new Error("Function not implemented.");
-}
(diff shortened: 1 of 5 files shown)


src/webchat-ui/components/presentational/DisconnectOverlay.tsx
```diff
@@ -1,78 +1,97 @@
...
+	position: "absolute",
+	left: 0,
+	top: 0,
+	width: "100%",
+	height: "100%",
 
-    zIndex: 2,
+	padding: 20,
+	boxSizing: "border-box",
+	backgroundColor: `color-mix(in srgb, ${theme.backgroundWebchat}, #0001)`,
```
dshire, line 34: I think we can just use theme.white here
backgroundWebchat is actually not the color for the chat history background, it's going on the header for example (see my open PR). I adjusted this according to the Figma designs (though they are a bit confusing in regards to usage of this color)

dshire (changes requested):
works fine. just one change I would suggest

dshire (approved):

PR #405, opened 2024-04-04
Feature/60603 fix v3 a11y issues

sushmi21:
Fix accessibility issues in webchat v3.

How to test:
- Please use your keyboard to navigate through the webchat and see if everything is accessible and focus is not lost
- check if focusable elements get a visible focus indication as expected. (Please comment below if you think we need to change focus styles of some elements, for example - input)
- See if the focus order is meaningful
NOTE: You will see some scrolling issues when reverse Navigating (SHIFT +TAB) through chat history when there are many messages. This has to be investigated and handled in a separate ticket

### DIFF
--- a/cypress/integration/collated-text-inputs.spec.ts
+++ b/cypress/integration/collated-text-inputs.spec.ts
@@ -125,31 +125,36 @@ describe("collated text inputs", () => {
         // TODO: make "collate" an opt-in on the "send message" api as an option and only use it on regular text inputs
         it("immediately sends a message triggered by a quick reply", () => {
             cy.withMessageFixture('quick-replies', () => {
-                cy.contains("foobar003qr01").click({force: true});
+                cy.get("button:contains('foobar003qr01')").last().focus();
+                cy.get("button:contains('foobar003qr01')").last().click({force: true});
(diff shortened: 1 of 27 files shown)


fabclj:
There is a UI issue on focus input text: there is a colored line on the bottom:
Screenshot from 2024-04-05 14-49-55

Are tests failing maybe for that?

sushmi21:
> There is a UI issue on focus input text: there is a colored line on the bottom: Screenshot from 2024-04-05 14-49-55
> 
> Are tests failing maybe for that?

This is not an issue. This is the focus indicator for the input. (Same as we have for v2). It looks odd because of the Branding text. I could not find a visually pleasing solution, so went with what we have in v2. I will discuss with Johanna to ask if she has better ideas

sushmi21:
> There is a UI issue on focus input text: there is a colored line on the bottom: Screenshot from 2024-04-05 14-49-55
> 
> Are tests failing maybe for that?

No, they are unrelated.

sushmi21:
> CR and functional test are done from my side. I see tests are still failing and PR from yesterday was success. Do you think are failing because this PR? If yes, we should fix them.

I think some of the tests are flaky. Different tests fail in different runs. We need do investigate and handle this separately

fabclj (approved):

PR #416, opened 2024-05-07
Update Cypress and Tests

kwinto:
This PR updates Cypress to the latest released version on v3. 

We also updated some test to be compatible with the new version. 2 tests were disabled. They will be fixed in a later PR.


cypress/e2e/collated-text-inputs.cy.ts
```diff
@@ -9,8 +9,8 @@ describe("collated text inputs", () => {
         cy.initMockWebchat()
             .openWebchat()
             .startConversation();
-        cy.get(".webchat-input-message-input").type("hi").type("{enter}");
-        cy.get(".webchat-input-message-input").type("whats up").type("{enter}");
+        cy.get(".webchat-input-message-input").wait(200).type("hi").type("{enter}");
```
sushmi21, line 12: Fabio investigated in the past and found that it is better to add typing delay instead of wait commands. Do you think we can use the delay command here?
Something like:
` cy.get(".webchat-input-message-input").type("whats up", { delay: 200 }).type("{enter}");`

sushmi21 (approved):

PR #418, opened 2024-05-10
Bug/61922 implement source color mapping

sushmi21:
Implement source color mapping for v3.

- Edit index.html to configure sourceColorMapping settings object inside widgetSettings
- Test this together with the PR in chat components repo
- Test by changing the values inside sourceColorMapping settings for different sources to 'primary', 'neutral' and 'default'


kwinto (changes requested):
"neutral" user message lacks bubble border.
"primary" bot message looks the same as neutral.

I am not sure if the requirements where discussed previously on this.

I would argue we should change the stylings a bit.

To me, we should give the users an opportunity to flip our main bubble styles (but without applying theme primary colour per se as its background).

The default for the user = currently used, slight background without border ="primary"
The default for the bot/agent = white background with a border = "neutral"

If the user sets, e.g { system: "primary" }, it should then use the same style which is used for the user messages (slight background, no border).

Let me know if you think we should clarify the requirements with PO. The older logic from version 2 does not combine here nicely.

sushmi21:
> "neutral" user message lacks bubble border. "primary" bot message looks the same as neutral.
> 
> I am not sure if the requirements where discussed previously on this.
> 
> I would argue we should change the stylings a bit.
> 
> To me, we should give the users an opportunity to flip our main bubble styles (but without applying theme primary colour per se as its background).
> 
> The default for the user = currently used, slight background without border ="primary" The default for the bot/agent = white background with a border = "neutral"
> 
> If the user sets, e.g { system: "primary" }, it should then use the same style which is used for the user messages (slight background, no border).
> 
> Let me know if you think we should clarify the requirements with PO. The older logic from version 2 does not combine here nicely.

I have made the fixes as discussed in the implementation ticket. I have also handled message bubble background and text color contract issue and the bubble border in this PR

sushmi21:
Tests will pass once the PR from chat-components is merged and released

kwinto (approved):

PR #422, opened 2024-05-22
Add minimize button, add reset logic to header close button

dshire:
https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/68601

Added minimization button. It has the functionality that the close button had, close button now also resets the webchat. To test, use the minimization button and compare it to the behavior of the close button from the different webchat components.


kwinto:
This adds a kind of weird UX to solve one very specific problem for one customer. Hope we can reconsider this in future and add an option instead.

kwinto (approved):

dshire:
> This adds a kind of weird UX to solve one very specific problem for one customer. Hope we can reconsider this in future and add an option instead.

I also don't love it, but I don't think it is bad UX as minimize and close buttons are what people are used to anyway

PR #423, opened 2024-05-23
Bug/61922 implement source direction mapping v3

sushmi21:
Test this along with https://github.com/Cognigy/chat-components/pull/60

- In index.html define the sourceDirectionMapping object.
- When a message source has the direction 'incoming', the bubble should be aligned to the left
- When a message source has the direction 'outcoming', the bubble should be aligned to the right
- Check the alignment of the typing indicator - should be aligned with the bot message direction

### DIFF
--- a/docs/embedding.md
+++ b/docs/embedding.md
@@ -546,7 +546,6 @@ interface IWebchatSettings {
 		sourceDirectionMapping: {
 			agent: TSourceDirection;
 			bot: TSourceDirection;
-			engagement: TSourceDirection;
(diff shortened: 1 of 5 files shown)


sushmi21:
Tests will pass once the chat-components PR is merged and released

sushmi21:
I think source direction mapping for engagement message does not make sense. I also do not see any changes in v2 for the engagement messages. Should I remove it from the config for v3? What do you think @kwinto ?

PR #425, opened 2024-05-27
Fix Datepicker default date formatting (v2)

fabclj:
This PR fixes a bug on DatePicker plugin where was not possible to submit the pre-selected defaultDate.

Work Item: [54236](https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/54236)

### DIFF
--- a/src/plugins/date-picker/index.tsx
+++ b/src/plugins/date-picker/index.tsx
@@ -312,9 +312,9 @@ const datePickerPlugin: MessagePluginFactory = ({ styled }) => {
       let locale = l10n[flatpickrLocaleId];
       const enableTime = !!data.enableTime;
 
-      const timeFormat = `${data.time_24hr ? "H" : "h"}:i${data.enableSeconds ? ":S" : ""}${
-      data.time_24hr ? "" : " K"
-    }`;
+      const timeTemp = data.time_24hr ? 'H:i' : 'h:i'; //12-hour format without AM/PM
+      const timeWithSeconds = data.enableSeconds ? `${timeTemp}:S` : timeTemp;
+      const timeFormat = data.time_24hr ? timeWithSeconds :`${timeWithSeconds} K` //12-hour format with AM/PM
 
       const dateFormatString = enableTime ? `${dateFormat} ${timeFormat}` : dateFormat;
       const dateFormatLocalString = `L${enableTime ? " LT" : ""}${data.enableSeconds ? "S" : ""}`;
--- a/src/plugins/date-picker/index.tsx
+++ b/src/plugins/date-picker/index.tsx
@@ -109,11 +109,11 @@ const datePickerPlugin: MessagePluginFactory = ({ styled }) => {
       borderColor: theme.greyColor,
       color: theme.greyColor,
       cursor: 'default'
-	},
-	'&:focus':{
-		outline: 'none',
-		boxShadow: `0 0 3px 1px ${theme.primaryWeakColor}`
-	}
+  },
+  '&:focus':{
+    outline: 'none',
+    boxShadow: `0 0 3px 1px ${theme.primaryWeakColor}`
+  }
   }));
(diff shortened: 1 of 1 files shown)


src/plugins/date-picker/index.tsx
```diff
@@ -307,9 +311,13 @@ const datePickerPlugin: MessagePluginFactory = ({ styled }) => {
       const flatpickrLocaleId = getFlatpickrLocaleId(localeId);
       let locale = l10n[flatpickrLocaleId];
       const enableTime = !!data.enableTime;
+
       const timeTemp = data.time_24hr ? 'H:i' : 'h:i'; //12-hour format without AM/PM
       const timeWithSeconds = data.enableSeconds ? `${timeTemp}:S` : timeTemp;
       const timeFormat = data.time_24hr ? timeWithSeconds :`${timeWithSeconds} K` //12-hour format with AM/PM
+
+      const dateFormatString = enableTime ? `${dateFormat} ${timeFormat}` : dateFormat;
+      const dateFormatLocalString = `L${enableTime ? " LT" : ""}${data.enableSeconds ? "S" : ""}`;
```
kwinto, line 320: We need to make sure to include the information about seconds-related change in changelog. 

I see the reason behind it, though we need to make it visible that the Webchat will now include seconds in the output if the setting enables them in the picker.

Please mention this in the PR description.

kwinto (changes requested):

src/plugins/date-picker/index.tsx
```diff
@@ -297,7 +301,7 @@ const datePickerPlugin: MessagePluginFactory = ({ styled }) => {
     static getOptionsFromMessage(message: IMessage) {
       const { data } = message.data._plugin;
 
-      const dateFormat = data.dateFormat || 'YYYY-MM-DD';
+      const dateFormat = data.dateFormat || 'Y-m-d';
```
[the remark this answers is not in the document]
kwinto: Any particular reason you see this is necessary?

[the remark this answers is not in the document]
fabclj: flatpickr format tokens are different from moment tokens.
https://flatpickr.js.org/formatting/#date-formatting-tokens
https://momentjs.com/docs/#/i18n/

this one 'YYYY-MM-DD' is for moment. 
this one is for 'Y-m-d'.

in this case we need the flatpickr one.

We are using flatpicker tokens also when we want to customize the node. We mention it in the docs:
https://docs.cognigy.com/ai/build/node-reference/data/datepicker/

By the way, we should extend the docs adding the tokens details for flatpicker

src/plugins/date-picker/index.tsx
fabclj: Basically, we had a bug in this part. We included the seconds when user enabled it and added a custom date format.
In case of no custom, we are localizing the date without seconds.

I agree, we need to include this fix on the changelog.

src/plugins/date-picker/index.tsx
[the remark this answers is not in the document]
fabclj: You can check this thing by passing this format tokens "YYYY-MM-DD" on the node editor level. You will see the date is broken. The correct default value need to be this one 'Y-m-d'.

src/plugins/date-picker/index.tsx
fabclj: I reopen this thread since I found a specific case to handle. I will fix in the next commit

fabclj: fixed

kwinto (approved):

src/plugins/date-picker/index.tsx
fabclj: For the records @kwinto:
After depper investigation and discussion we agreed on keep this part of code without any new change.

PR #435, opened 2024-07-02
Chat Options: TTS

kwinto:
This PR adds back TTS support to the V3. Also introduces relevant UI to the options screen.

We also try here to solve the ambiguity of having 2 settings for TTS activation:
`settings.behavior?.enableTTS` and `settings.chatOptions.activateTTSToggle`.

Besides that, we remove some dead code / build targets.

Success criteria:
- User can enable/disable TTS in Options.
- Messages are being read out when both mentioned above settings are active (or user sets the second option active from the options screen)

Deploy this AI branch for testing:
https://cognigy.visualstudio.com/Cognigy.AI/_git/cognigy/pullrequest/31024


sushmi21:
When TTS is not enabled, I see 2 section dividers in Chat options screen.
Screenshot from 2024-07-03 14-55-20

sushmi21:
What is the difference between enableTTS and activateTTSToggle setting? Setting true in one and false in the other seems contradicting. We should then at least add additional description in the embedding file and also in (Cognigy.AI) Endpoint editor field tooltips.

sushmi21:
When I enabled only enableTTS, incoming message was read out loud. Happened only once for me. Later none of the messages were read out, even if I set both settings to true.

kwinto:
> What is the difference between enableTTS and activateTTSToggle setting? Setting true in one and false in the other seems contradicting. We should then at least add additional description in the embedding file and also in (Cognigy.AI) Endpoint editor field tooltips.

The requirements to enable TTS are currently ambiguous. I have added additional info to the docs.

sushmi21 (approved):

PR #443, opened 2024-07-09
Add firefox cypress run

kwinto:


package.json
```diff
@@ -15,7 +15,9 @@
 		"cypress:open": "cypress open --e2e --browser chrome",
```
fabclj, line 15: I would suggest to add also "cypress:open:firefox" command here

PR #446, opened 2024-07-12
RTL direction fixes

kwinto:
This PR fixes some issues for the RTL locales.

# How to test
- open index.html,
-  add dir="rtl" attribute to the html tag
- link chat-components from this PR https://github.com/Cognigy/chat-components/pull/73
- open webchat
- check https://cognigy.visualstudio.com/Cognigy.AI/_workitems/edit/72592 for the success criteria 
- do a sanity check for thing which would be relevant for the RTL language support

### DIFF
--- a/.github/workflows/cypress-firefox.yml
+++ b/.github/workflows/cypress-firefox.yml
@@ -13,3 +13,9 @@ jobs:
         - run: npm ci
         - run: npm run build
         - run: npm run test:cypress:firefox
+        - name: Upload screenshots
+          uses: actions/upload-artifact@v4
+          if: failure()
+          with:
+            name: cypress-screenshots
+            path: cypress/screenshots
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -13,7 +13,7 @@ jobs:
         - run: npm ci
         - run: npm test
         - name: Upload screenshots
-          uses: actions/upload-artifact@v3
+          uses: actions/upload-artifact@v4
           if: failure()
           with:
             name: cypress-screenshots
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -12,3 +12,9 @@ jobs:
               cache: "npm"
         - run: npm ci
         - run: npm test
+        - name: Upload screenshots
+          uses: actions/upload-artifact@v3
+          if: failure()
+          with:
+            name: cypress-screenshots
+            path: cypress/screenshots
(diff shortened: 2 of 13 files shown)


sushmi21:
The chat toggle button itself appears in the left corner. I think this should not be affected.

Screenshot from 2024-07-18 16-29-27

sushmi21:
I remember we talked about switching the position of the icons. The icons are not still not switched. For example, the back button is to the right instead of left.
Screenshot from 2024-07-18 17-17-07

kwinto:
> I remember we talked about switching the position of the icons. The icons are not still not switched. For example, the back button is to the right instead of left. This applies to send button, close button etc. Screenshot from 2024-07-18 17-17-07

It is actually flipped. It is showing now to the "back direction" in RTL :-)

kwinto:
> The chat toggle button itself appears in the left corner. I think this should not be affected.
> 
> Screenshot from 2024-07-18 16-29-27

It should definitely appear on the left.

 In RTL localized websites, everything becomes “flipped” to usual for us direction, even the logotypes and such.

sushmi21:
> > I remember we talked about switching the position of the icons. The icons are not still not switched. For example, the back button is to the right instead of left. This applies to send button, close button etc. Screenshot from 2024-07-18 17-17-07
> 
> It is actually flipped. It is showing now to the "back direction" in RTL :-)

What I meant is the position of the buttons. The group of buttons like (minimize, close and chat options should be on the right) and the back button should be on the left, don't you think?

kwinto:
> > > I remember we talked about switching the position of the icons. The icons are not still not switched. For example, the back button is to the right instead of left. This applies to send button, close button etc. Screenshot from 2024-07-18 17-17-07
> > 
> > 
> > It is actually flipped. It is showing now to the "back direction" in RTL :-)
> 
> What I meant is the position of the buttons. The group of buttons like (minimize, close and chat options should be on the right) and the back button should be on the left, don't you think?

I think you got confused here. Originally, the close button is on the right. Now, in RTL, it is on the left. So we actually have flipped the position.

sushmi21:
> > > > I remember we talked about switching the position of the icons. The icons are not still not switched. For example, the back button is to the right instead of left. This applies to send button, close button etc. Screenshot from 2024-07-18 17-17-07
> > > 
> > > 
> > > It is actually flipped. It is showing now to the "back direction" in RTL :-)
> > 
> > 
> > What I meant is the position of the buttons. The group of buttons like (minimize, close and chat options should be on the right) and the back button should be on the left, don't you think?
> 
> I think you got confused here. Originally, the close button is on the right. Now, in RTL, it is on the left. So we actually have flipped the position.

No, I am not confused. I am saying that we should only flip the chat bubbles but not the icon positions. All icons should stay in their original position, don't you think?

sushmi21 (approved):

PR #453, opened 2024-07-30
Change readme text for v3

sushmi21:
This PR removes the usage warning for webchat v3 from the read-me.

Please let me know if we need to adjust the wordings here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
 # Cognigy Webchat Widget (v3)
 
-The `v3` branch contains a work-in-progress version of our next Webchat Widget and must not be used in production. We will announce our new Webchat Widget properly once it is ready.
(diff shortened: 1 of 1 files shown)


kwinto:
Maybe we should wait with the README update till we move v3 to the new repo. 

Then, we should add "deprecation notice" to this README for the v2 + add a link to the new repo. 
There will be no need to mention v3 branching in that case. 

@fabclj Agree?

fabclj:
> Maybe we should wait with the README update till we move v3 to the new repo.
> 
> Then, we should add "deprecation notice" to this README for the v2 + add a link to the new repo. There will be no need to mention v3 branching in that case.
> 
> @fabclj Agree?

Yes, I agree. In my understanding the plan is to remove v3 branch here after the new repo is ready and working.

sushmi21:
I agree too. This PR was created last week before the decision for new repo was made. I will abandon this Pr

PR #480, opened 2025-07-17
fix: #106042 update dom purify to fix regex vulenarability

vj-venkatesan:
Fixes [106042](https://cognigy.visualstudio.com/Boron/_workitems/edit/106042)
## Success Criteria
* Dom purify is udpated
* Santize should work as expected

## How to test
1. Chat with an endpoint with html content with insecure javascript in say nodes
2. Expect the messages to be sanitized and rendered correctly


vj-venkatesan:
https://github.com/Cognigy/WebchatWidget/pull/480/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55

Relevant change

